PR #11, opened 2016-07-14
Adjustments to match the introduction of AbstractMessageChannelBinder

mbogoevici:
Requires https://github.com/spring-cloud/spring-cloud-stream/pull/589 to be merged first.

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -22,21 +22,17 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.Map;
 import java.util.Properties;
 import java.util.UUID;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import kafka.admin.AdminUtils;
 import kafka.api.OffsetRequest;
 import kafka.api.TopicMetadata;
 import kafka.common.ErrorMapping;
-import kafka.serializer.Decoder;
 import kafka.serializer.DefaultDecoder;
 import kafka.utils.ZKStringSerializer$;
 import kafka.utils.ZkUtils;
@@ -51,25 +47,16 @@ import org.apache.kafka.common.utils.Utils;
 import scala.collection.Seq;
 
 import org.springframework.beans.factory.DisposableBean;
-import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
-import org.springframework.cloud.stream.binder.AbstractBinder;
+import org.springframework.cloud.stream.binder.AbstractMessageChannelBinder;
 import org.springframework.cloud.stream.binder.Binder;
 import org.springframework.cloud.stream.binder.BinderException;
 import org.springframework.cloud.stream.binder.BinderHeaders;
-import org.springframework.cloud.stream.binder.Binding;
-import org.springframework.cloud.stream.binder.DefaultBinding;
 import org.springframework.cloud.stream.binder.ExtendedConsumerProperties;
 import org.springframework.cloud.stream.binder.ExtendedProducerProperties;
 import org.springframework.cloud.stream.binder.ExtendedPropertiesBinder;
-import org.springframework.cloud.stream.binder.HeaderMode;
-import org.springframework.cloud.stream.binder.MessageValues;
-import org.springframework.cloud.stream.binder.PartitionHandler;
 import org.springframework.cloud.stream.binder.kafka.config.KafkaBinderConfigurationProperties;
-import org.springframework.http.MediaType;
-import org.springframework.integration.channel.FixedSubscriberChannel;
-import org.springframework.integration.endpoint.EventDrivenConsumer;
-import org.springframework.integration.handler.AbstractMessageHandler;
-import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
+import org.springframework.context.Lifecycle;
+import org.springframework.integration.core.MessageProducer;
 import org.springframework.integration.kafka.core.ConnectionFactory;
 import org.springframework.integration.kafka.core.DefaultConnectionFactory;
 import org.springframework.integration.kafka.core.KafkaMessage;
@@ -83,7 +70,7 @@ import org.springframework.integration.kafka.listener.KafkaMessageListenerContai
 import org.springframework.integration.kafka.listener.KafkaNativeOffsetManager;
 import org.springframework.integration.kafka.listener.MessageListener;
 import org.springframework.integration.kafka.listener.OffsetManager;
-import org.springframework.integration.kafka.support.KafkaHeaders;
+import org.springframework.integration.kafka.support.KafkaProducerContext;
 import org.springframework.integration.kafka.support.ProducerConfiguration;
 import org.springframework.integration.kafka.support.ProducerFactoryBean;
 import org.springframework.integration.kafka.support.ProducerListener;
@@ -92,9 +79,7 @@ import org.springframework.integration.kafka.support.ZookeeperConnect;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
-import org.springframework.messaging.MessageHeaders;
-import org.springframework.messaging.SubscribableChannel;
-import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.messaging.MessagingException;
 import org.springframework.retry.RetryCallback;
 import org.springframework.retry.RetryContext;
 import org.springframework.retry.RetryOperations;
@@ -109,7 +94,6 @@ import org.springframework.util.StringUtils;
 
 /**
  * A {@link Binder} that uses Kafka as the underlying middleware.
- *
  * @author Eric Bottard
  * @author Marius Bogoevici
  * @author Ilayaperumal Gopinathan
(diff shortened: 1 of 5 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -458,44 +265,73 @@ public Object doWithRetry(RetryContext context) throws RuntimeException {
...
+		this.topicsInUse.put(name, listenedPartitions);
+		return listenedPartitions;
+	}
 
-		final FixedSubscriberChannel bridge = new FixedSubscriberChannel(rh);
-		bridge.setBeanName("bridge." + name);
 
+	@Override
+	@SuppressWarnings("unchecked")
+	protected MessageProducer createConsumerEndpoint(String name, String group, Object queue,
```
markfisher: should the "queue" parameter be renamed "destination" (not just here but even more importantly in the superclass?)

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -458,44 +265,73 @@ public Object doWithRetry(RetryContext context) throws RuntimeException {
...
+				Executors.newFixedThreadPool(concurrency, DAEMON_THREAD_FACTORY);
 		final KafkaMessageListenerContainer messageListenerContainer = new KafkaMessageListenerContainer(
-				connectionFactory, listenedPartitions.toArray(new Partition[listenedPartitions.size()]));
+				this.connectionFactory, listenedPartitions.toArray(new Partition[listenedPartitions.size()])) {
 
-		if (logger.isDebugEnabled()) {
-			logger.debug("Listened partitions: " + StringUtils.collectionToCommaDelimitedString(listenedPartitions));
+			@Override
+			public void stop(Runnable callback) {
+				super.stop(callback);
```
markfisher: should the callback be invoked at the end of this method? (after the `dispatcherTaskExecutor` has been shutdown?)

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -549,76 +390,185 @@ public Object doWithRetry(RetryContext context) {
...
-						((DisposableBean) messageListenerContainer.getOffsetManager()).destroy();
+	@Override
+	protected void createProducerDestinationIfNecessary(String name,
+			ExtendedProducerProperties<KafkaProducerProperties> properties) {
+		if (this.logger.isInfoEnabled()) {
+			this.logger.info("Using kafka topic for outbound: " + name);
+		}
+		validateTopicName(name);
+		Collection<Partition> partitions = ensureTopicCreated(name, properties.getPartitionCount());
+		if (properties.getPartitionCount() < partitions.size()) {
```
markfisher: this logs a message, but in the line above the smaller count was passed to `ensureTopicCreated`... I admit I'm not fully familiar with the flow (yet), but wanted to point this out in case it's incorrect... not real clear when reading this code in any case - maybe the log message should be closer to where the other value supercedes

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -672,129 +623,57 @@ private String toDisplayString(String original, int maxCharacters) {
...
-				}
-				if (message.getPayload() instanceof byte[]) {
-					producerConfiguration.send(topicName, targetPartition, null, (byte[]) message.getPayload());
-				}
-				else {
-					logger.error("Raw mode supports only byte[] payloads but value sent was of type "
-							+ message.getPayload().getClass());
-				}
-			}
+		public void start() {
```
markfisher: I noticed `running` is initialized to `true` thus this is a no-op... assume that's intentional, but thought I'd point it out since it's slightly non-intuitive

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -59,8 +58,8 @@
 import org.springframework.retry.policy.SimpleRetryPolicy;
 import org.springframework.retry.support.RetryTemplate;
 
-import kafka.admin.AdminUtils;
-import kafka.api.TopicMetadata;
+import static org.assertj.core.api.Assertions.assertThat;
```
markfisher: static imports should be first

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -293,10 +293,11 @@ public void testCompression() throws Exception {
 		Arrays.fill(testPayload, (byte) 65);
 		KafkaTestBinder binder = getBinder();
 		for (ProducerMetadata.CompressionType codec : codecs) {
-			DirectChannel moduleOutputChannel = new DirectChannel();
-			QueueChannel moduleInputChannel = new QueueChannel();
 			ExtendedProducerProperties<KafkaProducerProperties> producerProperties = createProducerProperties();
 			producerProperties.getExtension().setCompressionType(codec);
+
+			DirectChannel moduleOutputChannel = createBindableChannel("input", createProducerBindingProperties(producerProperties));
```
markfisher: first param value should be "output"?

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
mbogoevici: No, this is right, I believe. We'd want to ensure that the superclass has closed cleanly before we stop the offset manager and shutdown the external task executor.

markfisher: If we need to ensure the superclass has closed, we should be passing a new callback to it (and waiting for it to be invoked), not the one that is passed here as a parameter (which is used by the application context when closing). The "callback" param should instead be invoked at the end of *this* method, so that the context does not think it's completed prematurely.

mbogoevici: Ok - although this is moot, as the call is always `stop()` with a null callback (from the Binding). We can clean this up.

markfisher: Ah, `null` is not very helpful here ;)

mbogoevici: Yeah - this is not actually done under the normal control of the context ...

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
mbogoevici: I will add a comment. It is possible for `ensureTopicCreated` to return a larger number of partitions even if we ask for a smaller number because a topic already exists (and has a larger number of partitions already, we never remove partitions - the parameter is a hint on the minimal number of partitions that the binder expects).

mbogoevici: If that is confusing, we should discuss the behaviour in a new issue.

mbogoevici: Perhaps we should at least document that it is a hint only.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
mbogoevici: We need to implement it because of Lifecycle. Perhaps we should implement Lifecycle properly instead (maybe without thread-safety)

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
mbogoevici: We observe the code conventions of Spring Boot ATM, so perhaps we can adjust this when we settle on a code convention set.

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
mbogoevici: Yeah, this points out how useless the parameter is.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
markfisher: A comment to clarify that would be fine.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/9cf07a3df41116ff07d281b81102842b65594bb3

PR #22, opened 2016-08-17
Use Spring Cloud Stream Parent directly

mbogoevici:
Since the intent for `spring-cloud-stream-dependencies` to
become the user-facing dependency management BOM, binder
implementations should switch to the Spring Cloud Stream
parent for version management.

This also ensures that they inherit the build lifecycle of
Spring Cloud Stream - e.g. checkstyle validation.


sobychacko:
LGTM - merged.

PR #39, opened 2016-09-12
Making AdminUtils optional

### LINKED ISSUE #36: Making AdminUtils and topic creation optional in Kafka binder
If an end user does not want to include the kafka broker jar in the classpath, we should support that model and not create the topic and simply ignore the fact that AdminUtils is not present. It shouldn't stop the binder.

sobychacko:
Resolves #36

### DIFF
--- a/spring-cloud-stream-binder-kafka-docs/src/main/asciidoc/overview.adoc
+++ b/spring-cloud-stream-binder-kafka-docs/src/main/asciidoc/overview.adoc
@@ -271,3 +271,32 @@ Then add these dependencies at the top of the `<dependencies>` section in the po
 The versions above are provided only for the sake of the example.
 For best results, we recommend using the most recent 0.10-compatible versions of the projects.
 ====
+
+==== Excluding Kafka broker jar from the classpath of the binder based application
+If the inclusion of the Kafka broker is not desired at runtime, Kafka binder allows this particular dependency to be excluded from the binder so that it will not be part of the classpath.
+However, topic creation will be disabled from the binder if this dependency is excluded.
+
+If you use Kafka 10 dependencies as advised above, all you have to do is not to include the kafka broker dependency.
+If you use Kafka 0.9, then ensure that you exclude the kafka broker jar from the `spring-cloud-starter-stream-kafka` dependency as following.
+
+[source,xml]
+----
+<dependency>
+  <groupId>org.springframework.cloud</groupId>
+  <artifactId>spring-cloud-starter-stream-kafka</artifactId>
+  <exclusions>
+    <exclusion>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka_2.11</artifactId>
+    </exclusion>
+  </exclusions>
+</dependency>
+----
+
+If you exclude the kafka broker dependency and the topic is not present on the server, then Kafka broker will create the topic if auto topic creation is enabled on the server.
+Please keep in mind that if you are relying on this, then the Kafka server will use the default number of partitions and replication factors.
+On the other hand, if auto topic creation is disabled on the server, then care must be taken before running the application to create the topic with the desired number of partitions.
+
+If you want to have full control over how partitions are allocated, then leave the default settings as they are, i.e. do not exclude the kafka broker jar and ensure that `spring.cloud.stream.kafka.binder.autoCreateTopics` is set to `true` which is the default.
+
+
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -198,8 +198,8 @@ public class KafkaMessageChannelBinder extends
 															ExtendedProducerProperties<KafkaProducerProperties> producerProperties) throws Exception {
 
 		KafkaTopicUtils.validateTopicName(destination);
-
-		Collection<PartitionInfo> partitions = ensureTopicCreated(destination, producerProperties.getPartitionCount());
+		createTopicsIfAutoCreateEnabledAndAdminUtilsPresent(destination, producerProperties.getPartitionCount());
+		Collection<PartitionInfo> partitions = getPartitionsForTopic(destination, producerProperties.getPartitionCount());
 
 		if (producerProperties.getPartitionCount() < partitions.size()) {
 			if (this.logger.isInfoEnabled()) {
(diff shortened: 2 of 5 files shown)


spring-cloud-stream-binder-kafka-docs/src/main/asciidoc/overview.adoc
```diff
@@ -271,3 +271,21 @@ Then add these dependencies at the top of the `<dependencies>` section in the po
 The versions above are provided only for the sake of the example.
 For best results, we recommend using the most recent 0.10-compatible versions of the projects.
 ====
+
+==== Excluding Kafka broker jar from the classpath of the binder based application
+If the inclusion of the Kafka broker is not desired at runtime, Kafka binder allows this particular dependency to be
+excluded from the binder so that it will not be part of the classpath. However, topic creation will be disabled from the binder
```
mbogoevici: Nit, but we use one sentence per line with asciidoc documentation - see http://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

spring-cloud-stream-binder-kafka-docs/src/main/asciidoc/overview.adoc
```diff
@@ -271,3 +271,21 @@ Then add these dependencies at the top of the `<dependencies>` section in the po
...
+if this dependency is excluded.
+
+If you use Kafka 10 dependencies as advised above, all you have to do is not to include the kafka broker dependency.
+If you use Kafka 0.9, then ensure that you exclude the kafka broker jar from the `spring-cloud-starter-stream-kafka` dependency.
```
mbogoevici: We could provide an explicit example here.

sobychacko: addressed the comments.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -431,35 +431,35 @@ private boolean isAutoCommitOnError(ExtendedConsumerProperties<KafkaConsumerProp
...
-				if (partitionSize < effectivePartitionCount) {
-					if (this.configurationProperties.isAutoAddPartitions()) {
-						adminUtilsOperation.invokeAddPartitions(zkUtils, topicName, effectivePartitionCount, null, false);
-					}
-					else {
-						throw new BinderException("The number of expected partitions was: " + partitionCount + ", but "
-								+ partitionSize + (partitionSize > 1 ? " have " : " has ") + "been found instead."
-								+ "Consider either increasing the partition count of the topic or enabling " +
-								"`autoAddPartitions`");
+		if (this.configurationProperties.isAutoCreateTopics() && adminUtilsOperation != null) {
```
mbogoevici: This could do with a bit of restructuring - at least renaming the method to something that indicates that it will happen only if topic autocreation is enabled. Actually wondering about moving the `this.configurationProperties.isAutoCreateTopics()` test outside the method. There's some minor duplication, but I think the code will read a bit clearer.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java
```diff
@@ -112,19 +114,14 @@ public AdminUtilsOperation kafka10AdminUtilsOperation() {
...
-	private static Method getMethod(ClassLoader classLoader, String methodName) {
-		try {
-			Class<?> adminUtilClass = classLoader.loadClass("kafka.admin.AdminUtils");
-			Method[] declaredMethods = adminUtilClass.getDeclaredMethods();
-			for (Method m : declaredMethods) {
-				if (m.getName().equals(methodName)) {
-					return m;
-				}
+	private static Method getMethod(ClassLoader classLoader, String methodName) throws ClassNotFoundException {
+		Class<?> adminUtilClass = classLoader.loadClass("kafka.admin.AdminUtils");
```
mbogoevici: We should use `org.springframework.util.ClassUtils#forName` with a `null` classloader so that we observe the thread context classloader if set (among other things).

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java
```diff
@@ -151,14 +154,20 @@ public boolean matches(ConditionContext conditionContext, AnnotatedTypeMetadata
...
+				Method addPartitions = getMethod(classLoader, "addPartitions");
+				if (addPartitions != null) {
+					Class<?>[] parameterTypes = addPartitions.getParameterTypes();
+					Class clazz = parameterTypes[parameterTypes.length - 1];
+					if (!clazz.getName().equals("kafka.admin.RackAwareMode")) {
+						return true;
+					}
 				}
 			}
+			catch (ClassNotFoundException e) {
```
mbogoevici: Might be clearer if we used a `@ConditionalOnClass`

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java
```diff
@@ -151,14 +154,20 @@ public boolean matches(ConditionContext conditionContext, AnnotatedTypeMetadata
 		public boolean matches(ConditionContext conditionContext, AnnotatedTypeMetadata annotatedTypeMetadata) {
 
 			ClassLoader classLoader = Kafka09Condition.class.getClassLoader();
-			Method addPartitions = getMethod(classLoader, "addPartitions");
-			if (addPartitions != null) {
-				Class<?>[] parameterTypes = addPartitions.getParameterTypes();
-				Class clazz = parameterTypes[parameterTypes.length - 1];
-				if (!clazz.getName().equals("kafka.admin.RackAwareMode")) {
-					return true;
+			try {
```
mbogoevici: One suggestion, missed when merged - but can we use `org.apache.kafka.common.utils.AppInfoParser#getVersion` to detect the Kafka client library version instead of the signature.

mbogoevici:
LGTM. Added a little polishing here: https://github.com/mbogoevici/spring-cloud-stream-binder-kafka/commit/002cd99c0ea7c3f30cb246d6e6a499f52f3585a1 . Let me know if you're ok with that.

PR #47, opened 2016-09-21
Tests/docs for auto commit offset off and manual acking. 

sobychacko:


spring-cloud-stream-binder-kafka-docs/src/main/asciidoc/overview.adoc
```diff
@@ -129,7 +129,10 @@ The property `spring.cloud.stream.instanceCount` must typically be greater than
 Default: `true`.
 autoCommitOffset::
   Whether to autocommit offsets when a message has been processed.
-If set to `false`, an `Acknowledgment` header will be available in the message headers for late acknowledgment.
+If set to `false`, an `org.springframework.kafka.support.Acknowledgment` header will be available in the message headers for late acknowledgment.
+In this case, the key for the header will be `org.springframework.kafka.support.KafkaHeaders.ACKNOWLEDGEMENT`.
```
mbogoevici: I think we should use the String value here.

mbogoevici:
Merged via https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/5f1e6caeab4adf304a5678700c07d86411199116

PR #57, opened 2016-10-11
Ability to override deserializer in Kafka consumer

### LINKED ISSUE #55: Add support for Avro messages encoded with Confluent's (de)serializer (and other alternative (de)serializers)
As described in http://stackoverflow.com/questions/39941497/spring-cloud-stream-kafka-consuming-avro-messages-from-confluent-rest-proxy, the Kafka Binders fail to deserialize Avro messages which have been serialized via [Confluents serializer](http://docs.confluent.io/3.0.0/app-development.html#java). The issue seems to be that the Confluent serializer adds 4 (or 5?) magic bytes at the beginning of each message which contains the Avro schema ID which can be used to retrieve the schema from the [schema registry](http://docs.confluent.io/3.0.0/schema-registry/docs/index.html).

There seems to be no easy way to configure the bindings to use alternative serializers. From what I understand, the bindings hard-code the use of the Kafka ByteArrayDeserializer (https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L254).

ilayaperumalg:
- When binding the consumer, the kafka consumer should not be set to use `ByteArrayDeserializer` for both key/value deserializer. Instead, they need to be used as the default values. Any extended consumer properties for key/value deserializer should override this default deserializer.
- Add test

This resolves #55

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -38,7 +38,6 @@ import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.security.JaasUtils;
 import org.apache.kafka.common.serialization.ByteArrayDeserializer;
 import org.apache.kafka.common.serialization.ByteArraySerializer;
-import org.apache.kafka.common.serialization.Deserializer;
 import org.apache.kafka.common.utils.Utils;
 
 import org.springframework.beans.factory.DisposableBean;
@@ -258,7 +257,6 @@ public class KafkaMessageChannelBinder extends
 				String.valueOf(producerProperties.getExtension().getBatchTimeout()));
 		props.put(ProducerConfig.COMPRESSION_TYPE_CONFIG,
 				producerProperties.getExtension().getCompressionType().toString());
-
 		if (!ObjectUtils.isEmpty(producerProperties.getExtension().getConfiguration())) {
 			props.putAll(producerProperties.getExtension().getConfiguration());
 		}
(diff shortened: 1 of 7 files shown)


spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -16,9 +16,14 @@
 
 package org.springframework.cloud.stream.binder.kafka;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+import static org.junit.Assert.*;
```
mbogoevici: Switch to individual imports.

mbogoevici: Use individual imports.

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -1187,6 +1198,46 @@ public void testPartitionCountNotReduced() throws Exception {
...
+	public void testConsumerDefaultDeserializer() throws Exception {
+		KafkaBinderConfigurationProperties configurationProperties = createConfigurationProperties();
+		DirectFieldAccessor consumerAccessor = new DirectFieldAccessor(getKafkaConsumer(configurationProperties));
+		assertTrue(consumerAccessor.getPropertyValue("keyDeserializer") instanceof  ByteArrayDeserializer);
+		assertTrue(consumerAccessor.getPropertyValue("valueDeserializer") instanceof  ByteArrayDeserializer);
+	}
+
+	@Test
+	@SuppressWarnings("unchecked")
+	public void testConsumerCustomDeserializer() throws Exception {
```
mbogoevici: It is not enough to test that overriding the properties works. We need to test that sending and receiving messages with these settings actually works.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -303,27 +302,18 @@ protected MessageProducer createConsumerEndpoint(String name, String group, Coll
...
-		Deserializer<byte[]> keyDecoder = new ByteArrayDeserializer();
-
 		if (!ObjectUtils.isEmpty(properties.getExtension().getConfiguration())) {
 			props.putAll(properties.getExtension().getConfiguration());
 		}
-
-		ConsumerFactory<byte[], byte[]> consumerFactory = new DefaultKafkaConsumerFactory<>(props, keyDecoder,
-				valueDecoder);
-
+		ConsumerFactory<byte[], byte[]> consumerFactory = new DefaultKafkaConsumerFactory<>(props);
```
mbogoevici: We should also support customizing the serialization on the consumer side.

mbogoevici:
Tried to run this, but the branch runs into test failures constantly (master runs fine) - can you check @ilayaperumalg ?

ilayaperumalg:
@mbogoevici Initially I thought this was because of the inconsistent test behaviour before. Now I realized it was a mistake on my part not to unbind some of the bindings created during the tests. Fixed them and updated the PR.

mbogoevici:
Cool - thanks @ilayaperumalg!

mbogoevici:
Rebased, squashed and merged via https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/36c90aa8c0acf1fe29e235b1d2aac05115fc4c90

PR #61, opened 2016-10-17
Use Spring Cloud Build as parent

### LINKED ISSUE #60: Do not use Spring Cloud Stream Parent as parent
Since version 1.1 we depend on Spring Cloud Stream as the parent for the build. This creates a tight coupling in what concerns the build cycle, etc. Should use Spring Cloud Build instead.

mbogoevici:
Fixes #60


ilayaperumalg:
Merged as 275f5cbb8edd0d12a8a13c10b6917450597efc74

PR #82, opened 2017-01-23
Embed exception info into DLQ forwarded messages

### LINKED ISSUE #32: Kafka DLQ forwarding should add error information in headers (if mode is not `raw`)
Copied from https://github.com/spring-cloud/spring-cloud-stream/issues/539

ilayaperumalg:
- When there are exceptions receiving the messages at the Kafka consumer, the rejected messages are embedded with exception messages with the header `exception`
   - The embedding happens only when the embed mode is set for headers
 - update test

This resolves #32

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -60,7 +60,6 @@ import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.integration.core.MessageProducer;
 import org.springframework.integration.kafka.inbound.KafkaMessageDrivenChannelAdapter;
 import org.springframework.integration.kafka.outbound.KafkaProducerMessageHandler;
-import org.springframework.integration.support.MessageBuilder;
 import org.springframework.kafka.core.ConsumerFactory;
 import org.springframework.kafka.core.DefaultKafkaConsumerFactory;
 import org.springframework.kafka.core.DefaultKafkaProducerFactory;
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -350,8 +355,24 @@ public void stop(Runnable callback) {
 				public void handle(Exception thrownException, final ConsumerRecord message) {
 					final byte[] key = message.key() != null ? Utils.toArray(ByteBuffer.wrap((byte[]) message.key()))
 							: null;
-					final byte[] payload = message.value() != null
+					byte[] payload = message.value() != null
 							? Utils.toArray(ByteBuffer.wrap((byte[]) message.value())) : null;
+					if (properties.getHeaderMode().equals(HeaderMode.embeddedHeaders)) {
+						EmbeddedHeadersMessageConverter embeddedHeadersMessageConverter = new EmbeddedHeadersMessageConverter();
+						try {
+							MessageValues messageValues = embeddedHeadersMessageConverter.extractHeaders(MessageBuilder.withPayload(payload).build(), true);
```
ilayaperumalg: Creation of `Message` can be avoided if we change the `EmbeddedHeaderMessageConverter` method signature like [this](https://github.com/ilayaperumalg/spring-cloud-stream/commit/6f22c99a11748a1a31b525c26af5b11589bfdd53)

mbogoevici: Wouldn't it be better if we overload the method instead of changing the entire signature, which could create compatibility problems?

ilayaperumalg:
Updated the PR based on https://github.com/spring-cloud/spring-cloud-stream/pull/781

ilayaperumalg:
Closing this to allow a better approach to handle this scenario as part of https://github.com/spring-cloud/spring-cloud-stream/issues/863

PR #84, opened 2017-01-24
Fixing race condition when topics are created in a stream and a TopicE…

### LINKED ISSUE #83: Sporadic `TopicExistsException` at start with 0.10.1.1 cluster
Using the Kafka binder with a 0.10.1.0 library version pointing at a 0.10.1.1. cluster results in startup failures with:

```
org.springframework.context.ApplicationContextException: Failed to start bean 'outputBindingLifecycle'; nested exception is org.apache.kafka.common.errors.TopicExistsExce
ption: Topic "ticktock.time" already exists.
```

Reproduced by both @trisberg and @sobychacko. 

Note: this could be the possible result of a race condition between two separate applications (producer and consumer) starting simultaneously as it has been observed only in Spring Cloud Data Flow deployments.

sobychacko:
…xistsException is thrown.

Fixes #83

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -223,7 +223,7 @@ public class KafkaMessageChannelBinder extends
 	protected String createProducerDestinationIfNecessary(String name,
 														ExtendedProducerProperties<KafkaProducerProperties> properties) {
 		if (this.logger.isInfoEnabled()) {
-			this.logger.info("Using - foobar kafka topic for outbound: " + name);
+			this.logger.info("Using - kafka topic for outbound: " + name);
 		}
 		KafkaTopicUtils.validateTopicName(name);
 		createTopicsIfAutoCreateEnabledAndAdminUtilsPresent(name, properties.getPartitionCount());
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -470,8 +470,18 @@ else if (errorCode == ErrorMapping.UnknownTopicOrPartitionCode()) {
...
 					public Object doWithRetry(RetryContext context) throws RuntimeException {
 
-						adminUtilsOperation.invokeCreateTopic(zkUtils, topicName, effectivePartitionCount,
-								configurationProperties.getReplicationFactor(), new Properties());
+						try {
+							adminUtilsOperation.invokeCreateTopic(zkUtils, topicName, effectivePartitionCount,
+									configurationProperties.getReplicationFactor(), new Properties());
+						}
+						catch (Exception e) {
+							if (e.getClass().getName().endsWith("TopicExistsException")){
```
mbogoevici: I think we should specifically check for the two possible exceptions here.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -470,8 +470,18 @@ else if (errorCode == ErrorMapping.UnknownTopicOrPartitionCode()) {
...
+								logger.warn("Attempted to create topic: " + topicName + ". Topic already exists.");
```
mbogoevici: We normally check if the `WARN` log level is enabled.

mbogoevici:
Squashed, merged via https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/c389fa3fa442632ff512946d04db6e46588a1f74

mbogoevici:
Merged to 1.1.x https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/tree/1.1.x

PR #91, opened 2017-02-07
Provisioning SPI related changes

### LINKED ISSUE #50: Kafka DLQ partitioning should match partition specifics of the original topic
As a developer, I'd like to make sure the DLQ partitioning mirrors the original Kafka partition specifications including partition-size, partition-count and the related configs. Also publish to the equivalent partition.

### LINKED ISSUE #86: Separate binding from queue provisioning
Downstream changes from https://github.com/spring-cloud/spring-cloud-stream/issues/745

sobychacko:
This PR depends on https://github.com/spring-cloud/spring-cloud-stream/pull/787

- Separate Kafka topic provisioning from the binder using the SPI provided in core
 - Refactor the common entities needed for Kafka into a new core module
 - Refactor binder code to reflect the SPI changes
 - Make DLQ provisioning match with partition properties of the topic

Fixes #86
Fixes #50


spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -284,7 +286,7 @@ public void testAutoCreateTopicsEnabledSucceeds() throws Exception {
 		FixedBackOffPolicy backOffPolicy = new FixedBackOffPolicy();
 		backOffPolicy.setBackOffPeriod(1000);
 		metatadataRetrievalRetryOperations.setBackOffPolicy(backOffPolicy);
-		setMetadataRetryOperations(binder, metatadataRetrievalRetryOperations);
+		//setMetadataRetryOperations(binder, metatadataRetrievalRetryOperations);
```
garyrussell: Why are these all commented out?

garyrussell: I just checked the binder and provisioner and they both now have a field for this data.

It looks like this code should be moved to a utility class if it's needed in both classes.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -92,65 +91,38 @@
...
+	public KafkaMessageChannelBinder(KafkaBinderConfigurationProperties configurationProperties,
+									KafkaTopicProvisioner provisioningProvider) {
+		super(false, headersToMap(configurationProperties), provisioningProvider);
+		this.configurationProperties = configurationProperties;
+		this.provisioningProvider = provisioningProvider;
 	}
 
 	/**
-	 * Retry configuration for operations such as validating topic creation
 	 *
```
garyrussell: We need to restore this Javadoc too - also, shouldn't we also pass them into the `KafkaTopicProvisioner` ??

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -92,65 +91,38 @@
...
-			String[] combinedHeadersToMap = Arrays.copyOfRange(BinderHeaders.STANDARD_HEADERS, 0,
-					BinderHeaders.STANDARD_HEADERS.length + configurationProperties.getHeaders().length);
-			System.arraycopy(configurationProperties.getHeaders(), 0, combinedHeadersToMap,
-					BinderHeaders.STANDARD_HEADERS.length,
-					configurationProperties.getHeaders().length);
-			headersToMap = combinedHeadersToMap;
-		}
-		return headersToMap;
-	}
+	KafkaTopicProvisioner provisioningProvider;
```
garyrussell: This field should be final (and moved to the top).

garyrussell: Actually - it's not used so why do we need it?

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
sobychacko: i removed that javadoc because topic creation validation is no longer the concern of the binder, but of the provisioner.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
sobychacko: will clean that up.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
garyrussell: But it's still used here, in `getPartitionsForTopic()`.

Maybe that should be moved to the provisioner too?

garyrussell: It would be a pain to have to set it in both objects.

mbogoevici: I think too that we should move `getPartitionsForTopic()` to the provisioner. Then the binder would have no code to interact with the metadata directly.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -194,49 +142,30 @@ public KafkaProducerProperties getExtendedProducerProperties(String channelName)
...
-	protected MessageHandler createProducerMessageHandler(final String destination,
+	protected MessageHandler createProducerMessageHandler(final ProducerDestination destination,
 															ExtendedProducerProperties<KafkaProducerProperties> producerProperties) throws Exception {
-
-		KafkaTopicUtils.validateTopicName(destination);
-		createTopicsIfAutoCreateEnabledAndAdminUtilsPresent(destination, producerProperties.getPartitionCount());
-		Collection<PartitionInfo> partitions = getPartitionsForTopic(destination, producerProperties.getPartitionCount());
-
+		final DefaultKafkaProducerFactory<byte[], byte[]> producerFB = getProducerFactory(producerProperties);
+		Collection<PartitionInfo> partitions = ((KafkaTopicProvisioner)provisioningProvider)
```
artembilan: That's what I mean in the https://github.com/spring-cloud/spring-cloud-stream/pull/804/files#r101065893.

As I said: just my fear.
If you are fine with this, we can go ahead.

garyrussell:
Merged as dc0bc18f3790322813ea3eda4ebb7c11c9a946ba

PR #93, opened 2017-02-08
Fix batchSize producer property

### LINKED ISSUE #92: Incorrect property name for producer batch size
Currently, the KafkaProducerProperties' `bufferSize` is set as `batchSize` in the Kafka producer configuration.

Due to backwards compatibility, not sure if this fix should go into 1.1.x as well.

ilayaperumalg:
- Update bufferSize -> batchSize
 - Add bufferMemory as producer property

Fixes #92


mbogoevici:
I don't think this is necessary, and in fact it is quite dangerous as it breaks backwards compatibility. Given that this is an alias for a configuration property, why can't we simply deprecated and point the user to the standard Kafka configuration?

ilayaperumalg:
that's why I had the question on backwards compatibility in the story. In any case, we can surely deprecate this property.

PR #95, opened 2017-02-13
Support Spring Boot KafkaProperties

### LINKED ISSUE #73: Support Boot 1.5 configuration properties for Kafka
Spring Boot 1.5 is introducing Kafka support, so we should be able to support the Kafka autoconfiguration the same way as we do with Rabbit.

ilayaperumalg:
- If KafkaProperties for the KafkaAutoConfiguration is set, then use those properties for the KafkaMessageChannelBinder
 - For the KafkaProperties that have explicit default, override with the KafkaMessageChannelBinder defaults when the properties are not set by any of the property sources
 - Support the existing Kafka Producer/Consumer properties if they are set
 - Add tests

Resolves #73

### DIFF
--- a/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaProducerProperties.java
+++ b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaProducerProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,8 +26,10 @@ import javax.validation.constraints.NotNull;
  */
 public class KafkaProducerProperties {
 
+	@Deprecated
 	private int bufferSize = 16384;
 
+	@Deprecated
 	private CompressionType compressionType = CompressionType.none;
 
 	private boolean sync;
@@ -80,6 +82,7 @@ public class KafkaProducerProperties {
 	public enum CompressionType {
 		none,
 		gzip,
-		snappy
+		snappy,
+		lz4
 	}
 }
--- a/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaProducerProperties.java
+++ b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaProducerProperties.java
@@ -27,8 +27,18 @@ import javax.validation.constraints.NotNull;
 public class KafkaProducerProperties {
 
 	@Deprecated
+	/**
+	 * bufferSize property is deprecated.
+	 * It is recommended to set compressionType as one of the per binding Kafka producer `configuration` properties.
+	 * If using KafkaAutoConfiguration from Spring Boot 1.5.x, `spring.kafka.producer.batchSize` property can also be used.
+	 */
 	private int bufferSize = 16384;
 
+	/**
+	 * compressionType property is deprecated.
+	 * It is recommended to set compressionType as one of the per binding Kafka producer `configuration` properties.
+	 * If using KafkaAutoConfiguration from Spring Boot 1.5.x, `spring.kafka.producer.compressionType` property can also be used.
+	 */
 	@Deprecated
 	private CompressionType compressionType = CompressionType.none;
 
@@ -79,6 +89,10 @@ public class KafkaProducerProperties {
 		this.configuration = configuration;
 	}
 
+	@Deprecated
+	/**
+	 * @see compressionType
+	 */
 	public enum CompressionType {
 		none,
 		gzip,
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
@@ -19,6 +19,9 @@ package org.springframework.cloud.stream.binder.kafka;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
+import org.apache.kafka.common.serialization.ByteArraySerializer;
+
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.env.EnvironmentPostProcessor;
 import org.springframework.core.env.ConfigurableEnvironment;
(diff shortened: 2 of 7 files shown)


ilayaperumalg:
Rebased.

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaProducerProperties.java
```diff
@@ -26,8 +26,10 @@
  */
 public class KafkaProducerProperties {
 
+	@Deprecated
```
sobychacko: It might be good to add a deprecation comment for what the suggested way is going forward.

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaProducerProperties.java
```diff
@@ -26,8 +26,10 @@
...
 	private int bufferSize = 16384;
 
+	@Deprecated
```
sobychacko: same as above.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -171,20 +178,27 @@ protected MessageHandler createProducerMessageHandler(final ProducerDestination
...
 		props.put(ProducerConfig.COMPRESSION_TYPE_CONFIG,
 				producerProperties.getExtension().getCompressionType().toString());
+		props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, this.configurationProperties.getKafkaConnectionString());
+		if (this.kafkaProperties != null) {
+			if (!this.kafkaProperties.getBootstrapServers().isEmpty()) {
+				props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, this.kafkaProperties.getBootstrapServers());
+			}
+			props.putAll(this.kafkaProperties.getProducer().buildProperties());
+			props.putAll(this.kafkaProperties.getProperties());
+		}
```
sobychacko: What happens if `kafkaProperties` is null (due to `spring-boot-autoconfigure` is not included in the application as this is an optional dependency), but the user sets a property `spring.kafka.producer.keySerializer`. I think in that case currently we ignore what the user sets and simply use `ByteArraySerializer`. Is that the intention? May be we need to add a doc saying that if any `spring.kafka.*` properties are used, then they must provide `spring-boot-autoconfigure`.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -308,20 +313,30 @@ public void onSuccess(SendResult<byte[], byte[]> result) {
...
 		props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, 100);
-		return props;
+		props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, this.configurationProperties.getKafkaConnectionString());
+		if (this.kafkaProperties != null) {
+			if (!this.kafkaProperties.getBootstrapServers().isEmpty()) {
+				props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, this.kafkaProperties.getBootstrapServers());
+			}
+			props.putAll(this.kafkaProperties.getConsumer().buildProperties());
+			props.putAll(this.kafkaProperties.getProperties());
+		}
```
sobychacko: same comment as above for the producer.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
ilayaperumalg: @sobychacko I think it is all about which Spring Boot version user provides. We go with Spring Boot `1.5.x` by default which enables Kafka auto configuration and if user explicitly overrides Spring Boot to `1.4.x` or before then KafkaAutoConfiguration wouldn't be available and there is no point of `spring.kafka.*` properties. Hence, it is not about just `spring-boot-autoconfigure` but about the Spring Boot version I believe.

sobychacko:
LGTM - squashed, polished - merging.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
```diff
@@ -19,26 +19,77 @@
...
-		propertiesToAdd.put("logging.level.kafka.admin.AdminClient.AdminConfig", "ERROR");
-		environment.getPropertySources().addLast(new MapPropertySource("kafkaBinderLogConfig", propertiesToAdd));
+		Map<String, Object> logProperties = new HashMap<>();
+		logProperties.put("logging.pattern.console", "%d{ISO8601} %5p %t %c{2}:%L - %m%n");
+		logProperties.put("logging.level.org.I0Itec.zkclient", "ERROR");
+		logProperties.put("logging.level.kafka.server.KafkaConfig", "ERROR");
+		logProperties.put("logging.level.kafka.admin.AdminClient.AdminConfig", "ERROR");
+		environment.getPropertySources().addLast(new MapPropertySource("kafkaBinderLogConfig", logProperties));
+		Map<String, Object> binderConfig = new HashMap<>();
+		if (environment.getProperty(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER) != null) {
```
mbogoevici: Why is this check necessary? We are simply populating a lowest-priority property source. Anything can override it.

mbogoevici: Second, this code runs on each refresh, so this would forcibly cast the value set to String, but below it is set to a class resulting in this kind of error at startup:
```
Caused by: org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [java.lang.Class<?>] to type [java.lang.String]
```

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
```diff
@@ -19,26 +19,77 @@
...
+			binderConfig.put(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER, environment.getProperty(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER));
+		}
+		else {
+			binderConfig.put(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER, ByteArraySerializer.class);
```
mbogoevici: Why are we setting the property value to a class instead of a class name? The property binding will know how to do the conversion, and elsewhere a String value is expected anyway.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
```diff
@@ -19,26 +19,77 @@
...
+		else {
+			binderConfig.put(SPRING_KAFKA_CONSUMER_VALUE_DESERIALIZER, ByteArrayDeserializer.class);
+		}
+		if (environment.getProperty(SPRING_KAFKA_BOOTSTRAP_SERVERS) != null) {
+			binderConfig.put(SPRING_KAFKA_BOOTSTRAP_SERVERS, environment.getProperty(SPRING_KAFKA_BOOTSTRAP_SERVERS));
+		}
+		else {
+			binderConfig.put(SPRING_KAFKA_BOOTSTRAP_SERVERS, "");
+		}
+		environment.getPropertySources().addLast(new MapPropertySource("kafkaBinderConfig", binderConfig));
```
mbogoevici: We can check if the property source already exists and simply not add it, in case this is running on a subsequent `refresh()` operation.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
```diff
@@ -19,26 +19,77 @@
...
+			binderConfig.put(SPRING_KAFKA_CONSUMER_VALUE_DESERIALIZER, environment.getProperty(SPRING_KAFKA_CONSUMER_VALUE_DESERIALIZER));
+		}
+		else {
+			binderConfig.put(SPRING_KAFKA_CONSUMER_VALUE_DESERIALIZER, ByteArrayDeserializer.class);
+		}
+		if (environment.getProperty(SPRING_KAFKA_BOOTSTRAP_SERVERS) != null) {
+			binderConfig.put(SPRING_KAFKA_BOOTSTRAP_SERVERS, environment.getProperty(SPRING_KAFKA_BOOTSTRAP_SERVERS));
+		}
+		else {
+			binderConfig.put(SPRING_KAFKA_BOOTSTRAP_SERVERS, "");
```
mbogoevici: Why do we populate an empty String explicitly here? The default of `localhost:9092` should be fine, I suppose?

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
ilayaperumalg: Yes, this needs to be of `String` type. will fix it shortly.

PR #99, opened 2017-02-17
Fix KafkaBinderEnvironmentPostProcessor

mbogoevici:
- Fixes for the way properties are registered
- Do not register the property source twice
- Bootstrap test

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
@@ -26,6 +26,7 @@ import org.springframework.boot.SpringApplication;
 import org.springframework.boot.env.EnvironmentPostProcessor;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.MapPropertySource;
+import org.springframework.core.env.PropertySource;
 
 /**
  * An {@link EnvironmentPostProcessor} that sets some common configuration properties (log config etc.,) for Kafka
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
```diff
@@ -49,47 +50,30 @@
...
-			binderConfig.put(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER, environment.getProperty(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER));
+		if (!hasPropertySource(KAFKA_BINDER_DEFAULT_PROPERTIES, environment)) {
+			Map<String, Object> kafkaBinderDefaultProperties = new HashMap<>();
+			kafkaBinderDefaultProperties.put("logging.pattern.console", "%d{ISO8601} %5p %t %c{2}:%L - %m%n");
+			kafkaBinderDefaultProperties.put("logging.level.org.I0Itec.zkclient", "ERROR");
+			kafkaBinderDefaultProperties.put("logging.level.kafka.server.KafkaConfig", "ERROR");
+			kafkaBinderDefaultProperties.put("logging.level.kafka.admin.AdminClient.AdminConfig", "ERROR");
+			kafkaBinderDefaultProperties.put(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER, ByteArraySerializer.class.getName());
+			kafkaBinderDefaultProperties.put(SPRING_KAFKA_PRODUCER_VALUE_SERIALIZER, ByteArraySerializer.class.getName());
+			kafkaBinderDefaultProperties.put(SPRING_KAFKA_CONSUMER_KEY_DESERIALIZER, ByteArrayDeserializer.class);
```
ilayaperumalg: These `deserializer` properties are also supposed to be of `String` type.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderEnvironmentPostProcessor.java
```diff
@@ -49,47 +50,30 @@
...
-		Map<String, Object> binderConfig = new HashMap<>();
-		if (environment.getProperty(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER) != null) {
-			binderConfig.put(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER, environment.getProperty(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER));
+		if (!hasPropertySource(KAFKA_BINDER_DEFAULT_PROPERTIES, environment)) {
+			Map<String, Object> kafkaBinderDefaultProperties = new HashMap<>();
+			kafkaBinderDefaultProperties.put("logging.pattern.console", "%d{ISO8601} %5p %t %c{2}:%L - %m%n");
+			kafkaBinderDefaultProperties.put("logging.level.org.I0Itec.zkclient", "ERROR");
+			kafkaBinderDefaultProperties.put("logging.level.kafka.server.KafkaConfig", "ERROR");
+			kafkaBinderDefaultProperties.put("logging.level.kafka.admin.AdminClient.AdminConfig", "ERROR");
+			kafkaBinderDefaultProperties.put(SPRING_KAFKA_PRODUCER_KEY_SERIALIZER, ByteArraySerializer.class.getName());
```
ilayaperumalg: I think we need to check if these `serialization` and `deserialization` properties to see if they are set by the user explicitly. That was the intention for having the non-null check originally in the EnvironmentPostProcessor. This way, if the user wants to go with `native` serialization/de-serialization support, then those properties would never be set. or, am I missing something here?

mbogoevici: We don't need to check because every other property source in which the user might have set them would have a higher priority than the one we are just adding.

mbogoevici: This is actually tested by `KafkaBinderConfigurationTest` I guess :)

ilayaperumalg: ok :-) I will check that out.

mbogoevici:
Closing this PR until we revise the 1.5 property changes.

PR #100, opened 2017-02-17
Consolidate Rawmode tests as part of KafkaBinderTests

### LINKED ISSUE #40: Add Raw mode Kafka tests for 0.10 tests.

sobychacko:
Enable all version specific Kafka tests to run raw mode tests as well.

This way we don't need to maintain a separate battery of raw mode tests and as long as the main tests are extended from `KafkaBinderTests`, they get the benefit of raw mode tests being run. 

Fixes #40


mbogoevici:
Merged via https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/2530229fb52eaf89556b4f21f09aadd2ef229f27

PR #105, opened 2017-02-20
Remove JLine as dependency

mbogoevici:
This is the PR for the 1.1.x branch.


mbogoevici:
Actually, the dependency must be removed in the parent.

PR #107, opened 2017-02-21
Support KafkaProperties from Spring Boot autoconfiguration

### LINKED ISSUE #73: Support Boot 1.5 configuration properties for Kafka

ilayaperumalg:
- If KafkaProperties is available from KafkaAutoConfiguration, retrieve the kafka properties and set as `KafkaBinderConfigurationProperties`' configuration property.
This way, these properties are available at the top level and per-binding properties could still override when setting producer/consumer properties during binding operation
 - Add tests to verify the scenarios

Resolves #73


sobychacko:
Reviewed, polished and squashed the commits. Merging.

PR #119, opened 2017-04-03
disable parititon sanity check if auto rebalancing

sflandergan:


pivotal-issuemaster:
@sflandergan Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream-binder-kafka&pullRequestId=119)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-stream-binder-kafka&pullRequestId=119) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@sflandergan Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream-binder-kafka&pullRequestId=119)!

sobychacko:
Instead of hardcoding the `partitionCount` to be 1 when `autoRebalanceEnabled` is true, why don't we control this logic through a boolean flag, something like `allowIdleConsumers`. If its set to `true`, we don't do this check and if `false`, we follow the strict rule of not allowing less number of partitions than consumers (thereby not allowing idle consumers). In Kafka 0.8, this check was necessary as no auto rebalancing was available there. 

Please keep in mind that, Chelsea.RELEASE is imminent and therefore we have to defer this change to either SR1 or another future release.

sflandergan:
@sobychacko up to you, I just wanted to make a minimal invasive change ;)

sobychacko:
@sflandergan I think we can certainly iterate on a change. Feel free to update the PR.

sflandergan:
@sobychacko I've provided you an alternative with different handling options for the unexpected partition count. This also covers issues within the administration part of topics.

Cheers
Simon

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
```diff
@@ -231,7 +230,10 @@ public Object doWithRetry(RetryContext context) throws RuntimeException {
 		}
 	}
 
-	public Collection<PartitionInfo> getPartitionsForTopic(final int partitionCount, final Callable<Collection<PartitionInfo>> callable) {
+	public Collection<PartitionInfo> getPartitionsForTopic(final int partitionCount, 
+			UnexpectedPartitionCountHandling partitionCountExpectationHandling,
```
sobychacko: Compile errors. This has to be declared as final as an inner class is accessing this variable.

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
```diff
@@ -326,5 +327,48 @@ public String toString() {
...
+			}
+		};
+	}
+	
+	public UnexpectedPartitionCountHandling producerHandling() {
+		return new UnexpectedPartitionCountHandling() {
+
+			@Override
+			public void handlePartitionCountTooLow(String topicName, int partitionSize, int effectivePartitionCount) {
+				logger.debug(String.format("There is only %s partitions on topic %s, but producer has % partitions configured",
```
sobychacko: Missing format specifier.

sobychacko:
@sflandergan thank you for the PR. 
Some general comments about the design. In the provisioner, currently, we throw an exception and fail-fast if the number of partitions the user provides is greater than the partitions present for an existing topic. This is true currently for both producer and consumer. The PR changes this logic slightly so that we only throw the exception and fail fast selectively, in particular when auto rebalancing is **not** enabled. When auto rebalancing is enabled, we no longer throw this exception. Producer provisioning is also changed in that we don't fail fast in this scenario, but simply log a message. For consumer, I see the value in the changes. However, for producer, we might want to think this through a bit more as we are changing the behavior there. Lets run through this from a producer provisioning standpoint.  Producer `A` wants to provision topic `foo` with `4` partitions. However, `foo` already exists and it has only `2` partitions on the broker. If the `autoAddPartitions` is enabled, then this is fine. We will initiate a change in the number of partitions, but if the user forgets to enable this, with this PR, the user won't know that there was an option to change the number of partitions. By fail-fasting as it does today, it gives the user that opportunity to enable that property.

sflandergan:
@sobychacko I reverted to the producer handling to the initial behavior. I didn't think of the case where applications might want to send to specific partitions and want to ensure, that these partitions are also available. 

I think with the changed design you can easily adapt handling of unexpected partition counts.

Cheers
Simon

sobychacko:
@sflandergan I am still getting a compile error on the build. How do you build the project? 
Do the following at the root of the project: 
`./mvnw clean package` - this fails for me. 

Thanks!

sflandergan:
@sobychacko should work now. I also rebased from master branch, since the binder tests were failing.

sobychacko:
@sflandergan looks good. Than you for the contribution. Added author, polished and merged. We might have to do a bit more refactoring, but will do it on a separate PR.

PR #125, opened 2017-04-17
SCSt-GH-913: Error Handling via ErrorChannel

garyrussell:
- configure an ErrorMessageSendingRecoverer to send errors to an error channel, whether or not retry is enabled.

__Needs supporting PRs from SCSt, Spring Integration Kafka, Spring Integration__

Depends on spring-cloud/spring-cloud-stream#913.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,8 +12,8 @@
 	</parent>
 	<properties>
 		<java.version>1.7</java.version>
-		<kafka.version>0.10.2.1</kafka.version>
-		<spring-kafka.version>1.2.0.RELEASE</spring-kafka.version>
+		<kafka.version>0.10.1.1</kafka.version>
+		<spring-kafka.version>1.1.6.RELEASE</spring-kafka.version>
 		<spring-integration-kafka.version>2.1.1.RELEASE</spring-integration-kafka.version>
 		<spring-cloud-stream.version>1.3.0.BUILD-SNAPSHOT</spring-cloud-stream.version>
 	</properties>
--- a/pom.xml
+++ b/pom.xml
@@ -12,9 +12,9 @@
 	</parent>
 	<properties>
 		<java.version>1.7</java.version>
-		<kafka.version>0.10.1.1</kafka.version>
-		<spring-kafka.version>1.1.2.RELEASE</spring-kafka.version>
-		<spring-integration-kafka.version>2.1.0.RELEASE</spring-integration-kafka.version>
+		<kafka.version>0.10.2.1</kafka.version>
+		<spring-kafka.version>1.2.0.RELEASE</spring-kafka.version>
+		<spring-integration-kafka.version>2.1.1.RELEASE</spring-integration-kafka.version>
 		<spring-cloud-stream.version>1.3.0.BUILD-SNAPSHOT</spring-cloud-stream.version>
 	</properties>
 	<modules>
@@ -92,7 +92,7 @@
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
-	
+
 	<build>
 		<pluginManagement>
 			<plugins>
--- a/spring-cloud-stream-binder-kafka/pom.xml
+++ b/spring-cloud-stream-binder-kafka/pom.xml
@@ -27,17 +27,6 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-stream</artifactId>
-			<version>1.3.0.BUILD-SNAPSHOT</version>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.integration</groupId>
-			<artifactId>spring-integration-core</artifactId>
-			<version>4.3.10.BUILD-SNAPSHOT</version>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.integration</groupId>
-			<artifactId>spring-integration-jmx</artifactId>
-			<version>4.3.10.BUILD-SNAPSHOT</version>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
--- a/spring-cloud-stream-binder-kafka/pom.xml
+++ b/spring-cloud-stream-binder-kafka/pom.xml
@@ -27,6 +27,17 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-stream</artifactId>
+			<version>1.3.0.BUILD-SNAPSHOT</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-core</artifactId>
+			<version>4.3.10.BUILD-SNAPSHOT</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-jmx</artifactId>
+			<version>4.3.10.BUILD-SNAPSHOT</version>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
@@ -58,7 +69,7 @@
 		<dependency>
 			<groupId>org.springframework.integration</groupId>
 			<artifactId>spring-integration-kafka</artifactId>
-			<version>${spring-integration-kafka.version}</version>
+			<version>2.1.1.BUILD-SNAPSHOT</version>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -52,20 +52,22 @@ import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.integration.core.MessageProducer;
 import org.springframework.integration.kafka.inbound.KafkaMessageDrivenChannelAdapter;
 import org.springframework.integration.kafka.outbound.KafkaProducerMessageHandler;
+import org.springframework.integration.kafka.support.RawRecordHeaderErrorMessageStrategy;
+import org.springframework.integration.support.ErrorMessageStrategy;
 import org.springframework.kafka.core.ConsumerFactory;
 import org.springframework.kafka.core.DefaultKafkaConsumerFactory;
 import org.springframework.kafka.core.DefaultKafkaProducerFactory;
 import org.springframework.kafka.core.KafkaTemplate;
 import org.springframework.kafka.listener.AbstractMessageListenerContainer;
 import org.springframework.kafka.listener.ConcurrentMessageListenerContainer;
-import org.springframework.kafka.listener.ErrorHandler;
 import org.springframework.kafka.listener.config.ContainerProperties;
 import org.springframework.kafka.support.ProducerListener;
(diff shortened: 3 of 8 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -252,9 +256,11 @@ public void stop(Runnable callback) {
 					}
 				};
 		messageListenerContainer.setConcurrency(concurrency);
-		messageListenerContainer.getContainerProperties().setAckOnError(isAutoCommitOnError(extendedConsumerProperties));
+		messageListenerContainer.getContainerProperties()
+				.setAckOnError(isAutoCommitOnError(extendedConsumerProperties));
```
artembilan: Not related to this change, but since I have noticed the fact, I decided to comment.

There is a PR in Spring Kafka https://github.com/spring-projects/spring-kafka/pull/293 which purpose to disable `ackOnError` if the mode is MANUAL.
Something similar is said on the appropriate JavaDocs:
```
* Set whether or not the container should commit offsets (ack messages) where the
 * listener throws exceptions. This works in conjunction with {@link #ackMode} and is
 * effective only when the kafka property {@code enable.auto.commit} is {@code false};
 * it is not applicable to manual ack modes. 
```

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -268,29 +274,40 @@ public void stop(Runnable callback) {
 				new KafkaMessageDrivenChannelAdapter<>(
 						messageListenerContainer);
 		kafkaMessageDrivenChannelAdapter.setBeanFactory(this.getBeanFactory());
-		final RetryTemplate retryTemplate = buildRetryTemplate(extendedConsumerProperties);
-		kafkaMessageDrivenChannelAdapter.setRetryTemplate(retryTemplate);
+		if (extendedConsumerProperties.getMaxAttempts() > 1) {
```
mbogoevici: Wondering if in a next pass we could move all up to `AbstractMessageChannelBinder` assuming that the ChannelAdapter used for a binder accepts a recoveryCallback and an ErrorChannel ?

garyrussell:
Reworked for the [latest abstract binder changes](https://github.com/spring-cloud/spring-cloud-stream/pull/926).

Before moving the retry stuff up to the abstract binder, we would need s-i-amqp changes - it wires retry a completely different way (it goes into the listener container, not the channel adapter). I was hoping to avoid changing that. 

I'll start taking a look over there (rabbit binder), if I need other s-i-amqp changes, we can revisit this.

spring-cloud-stream-binder-kafka/pom.xml
```diff
@@ -58,7 +69,7 @@
 		<dependency>
 			<groupId>org.springframework.integration</groupId>
 			<artifactId>spring-integration-kafka</artifactId>
-			<version>${spring-integration-kafka.version}</version>
+			<version>2.1.1.BUILD-SNAPSHOT</version>
```
artembilan: Looks like `<spring-integration-kafka.version>2.1.0.RELEASE</spring-integration-kafka.version>` can be used from the root POM. 
Why do you do that explicitly here?
Thanks

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -305,11 +341,12 @@ public void onSuccess(SendResult<byte[], byte[]> result) {
 										"Sent to DLQ " + sb.toString());
```
artembilan: There is similar log message not visible here on GH: `"Error sending to DLQ" + sb.toString()`.
Must contains space after `DLQ` word.

spring-cloud-stream-binder-kafka/pom.xml
garyrussell: As I allude to in the PR description; it needs https://github.com/spring-projects/spring-integration-kafka/pull/163

spring-cloud-stream-binder-kafka/pom.xml
```diff
@@ -27,6 +27,17 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-core</artifactId>
+			<version>4.3.10.BUILD-SNAPSHOT</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-jmx</artifactId>
+			<version>4.3.10.BUILD-SNAPSHOT</version>
```
garyrussell: Just overriding all of the transitive SI deps so that they are the same version.

spring-cloud-stream-binder-kafka/pom.xml
artembilan: Right, that's correct. The main question here why don't do that (`2.1.1.BUILD-SNAPSHOT`) in the root pom, but just only here in the target project?
Feels like inconsistency.

garyrussell: This stuff will all be sorted out when boot 1.5.4 is released. I have scheduled 2.1.1 for tomorrow.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -272,34 +280,57 @@ public void stop(Runnable callback) {
...
+		}
+		else {
+			kafkaMessageDrivenChannelAdapter.setErrorChannel(errorInfrastructure.getErrorChannel());
+		}
+		return kafkaMessageDrivenChannelAdapter;
+	}
+
+	@Override
+	protected ErrorMessageStrategy getErrorMessageStrategy() {
+		return new RawRecordHeaderErrorMessageStrategy();
```
mbogoevici: Should we perhaps use a different strategy here - it could be beneficial to get the original exception as well.

garyrussell: ??

All this does is add the raw `ConsumerRecord` as as header to the `ErrorMessage`.

The `ErrorMessage` has a payload of a `MessagingException` that has the `failedMessage` (`spring-messaging` `Message<?>`) and `cause` properties.

mbogoevici: Ouch. My mistake, sorry. I got confused about what gets passed to the `errorMessageHandler` below.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -272,34 +280,57 @@ public void stop(Runnable callback) {
...
+					final ConsumerRecord<?, ?> record = message.getHeaders()
+							.get(KafkaMessageDrivenChannelAdapter.KAFKA_RAW_DATA, ConsumerRecord.class);
+					final byte[] key = record.key() != null ? Utils.toArray(ByteBuffer.wrap((byte[]) record.key()))
 							: null;
-					final byte[] payload = message.value() != null
-							? Utils.toArray(ByteBuffer.wrap((byte[]) message.value())) : null;
-					String dlqName = StringUtils.hasText(extendedConsumerProperties.getExtension().getDlqName()) ?
-							extendedConsumerProperties.getExtension().getDlqName() : "error." + destination.getName() + "." + group;
-					ListenableFuture<SendResult<byte[], byte[]>> sentDlq = kafkaTemplate.send(dlqName, message.partition(), key, payload);
+					final byte[] payload = record.value() != null
```
mbogoevici: Wondering if it makes sense to manipulate the payload so that we can insert the error message into the headers - i.e. extract the headers, add the exception, and re-encode the payload with the exception stacktrace and message as it is done for Rabbit.

spring-cloud-stream-binder-kafka/pom.xml
artembilan: I think this PR need some versions update as all others similar.
I mean that we don't need this `4.3.10.BUILD-SNAPSHOT`, the GA is polled by SCSt transitively.
Seems for me the same is for SIK

spring-cloud-stream-binder-kafka/pom.xml
artembilan: BTW each testing module has its own:
```
<spring-kafka.version>1.1.2.RELEASE</spring-kafka.version>
<spring-integration-kafka.version>2.1.0.RELEASE</spring-integration-kafka.version>
```
So, I end up with the `ClassNotFound` when run tests.

garyrussell:
>I think this PR need some versions update as all others similar.

Right; as I mentioned in the standup meeting; I am waiting to get the rabbit binder PR merged before finishing this one off, rather than keeping bouncing back and forth between them.

artembilan:
Ok, sorry, I didn't get that properly then. We need SCSt PR merged first, but I don't feel confident in that. Would great to get some :+1: from some one closer to SCSt. 
Then I'll merge Rabbit PR with appropriate versions polishing and we will be ready nail this one finally.

pom.xml
```diff
@@ -12,16 +12,15 @@
 	</parent>
 	<properties>
 		<java.version>1.7</java.version>
-		<kafka.version>0.10.1.1</kafka.version>
-		<spring-kafka.version>1.1.2.RELEASE</spring-kafka.version>
-		<spring-integration-kafka.version>2.1.0.RELEASE</spring-integration-kafka.version>
+		<kafka.version>0.10.2.1</kafka.version>
+		<spring-kafka.version>1.2.0.RELEASE</spring-kafka.version>
```
mbogoevici: This a bit of an issue if it is a hard requirement, as applications created with initializr/Boot 1.5.4 will not get this dependency. Based on the fact that Spring Integration stays at 2.1.x I suppose that we can revert to 1.1.x ?

mbogoevici: A cursory test shows that this works with the previous versions (and we do test with Kafka 0.10.2 separately). So we may just want to put them back - not the SI-Kafka version though.

spring-cloud-stream-binder-kafka/pom.xml
mbogoevici: 4.3.10 is now brought in by Boot 1.5.4.

pom.xml
```diff
@@ -13,8 +13,8 @@
 	<properties>
 		<java.version>1.7</java.version>
 		<kafka.version>0.10.1.1</kafka.version>
-		<spring-kafka.version>1.1.2.RELEASE</spring-kafka.version>
-		<spring-integration-kafka.version>2.1.0.RELEASE</spring-integration-kafka.version>
+		<spring-kafka.version>1.1.6.RELEASE</spring-kafka.version>
```
mbogoevici: Maybe in the future we can remove these entirely, as the baseline is now Boot 1.5

mbogoevici:
Closed via https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/07e15c141e912c0f27a5a30650ba88f3c043a2fe

PR #127, opened 2017-04-19
Confluent repository needed for tests is moved

### LINKED ISSUE #121: Kafka Binder POM contains repository info
We shouldn't specify a direct repository reference in the POM:

```
https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka/pom.xml#L104
```

While this is currently not likely to affect users as the only artifacts in use are in test scope, this is still a bad practice and we should avoid it. 

We can make a separate artifact for tests that depend on that repository.

sobychacko:
Confluent repository needed for tests is moved to spring-cloud-stream-binder-kafka-test-support

Following dependencies are moved to spring-cloud-stream-binder-kafka-test-support

 *spring-cloud-stream-schema
 *kafka-avro-serializer
 *kafka-schema-registry

spring-cloud-stream-binder-kafka-test-support is brought into spring-cloud-stream-binder-kafka only in test scope

Fixes #121


mbogoevici:
I will need to test this, but I think that there's still some kind of transitive dependency between `spring-cloud-stream-binder-kafka` and the Confluent artifacts (still in test scope). Can we perhaps make a test artifact inside the project that depends on `spring-cloud-stream-binder-kafka` and contains those tests ?

sobychacko:
@mbogoevici another set of commit is pushed to the PR.

mbogoevici:
I suggest we postpone merging this until we have a 1.3 branch later this week.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as d4aaf78

PR #143, opened 2017-05-22
Added Kafka binder consumer lag metrics

hekonsek:
Hi,

I have created a simple metric indicating what is a current lag in a consumption of the Kafka messages for binding consumer. With this metric users can easily see if given microservice is too slow in consuming messages and should be scaled horizontally for example. I plan to use this metric in order to enable auto-scaling in Kubernetes for Spring Cloud Stream based microservices.

What do you think?

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,6 +32,7 @@ import org.springframework.kafka.core.ConsumerFactory;
  *
  * @author Ilayaperumal Gopinathan
  * @author Marius Bogoevici
+ * @author Henryk Konsek
  */
 public class KafkaBinderHealthIndicator implements HealthIndicator {
 
@@ -53,7 +54,7 @@ public class KafkaBinderHealthIndicator implements HealthIndicator {
 			for (String topic : this.binder.getTopicsInUse().keySet()) {
 				List<PartitionInfo> partitionInfos = metadataConsumer.partitionsFor(topic);
 				for (PartitionInfo partitionInfo : partitionInfos) {
-					if (this.binder.getTopicsInUse().get(topic).contains(partitionInfo) && partitionInfo.leader()
+					if (this.binder.getTopicsInUse().get(topic).getPartitionInfos().contains(partitionInfo) && partitionInfo.leader()
 							.id() == -1) {
 						downMessages.add(partitionInfo.toString());
 					}
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
@@ -16,19 +16,28 @@
 package org.springframework.cloud.stream.binder.kafka;
 
 import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.boot.actuate.endpoint.PublicMetrics;
 import org.springframework.boot.actuate.metrics.Metric;
+import org.springframework.cloud.stream.binder.kafka.properties.KafkaBinderConfigurationProperties;
+import org.springframework.cloud.stream.config.BindingProperties;
+import org.springframework.cloud.stream.config.BindingServiceProperties;
+import org.springframework.core.env.PropertyResolver;
 import org.springframework.kafka.core.ConsumerFactory;
+import org.springframework.kafka.core.DefaultKafkaConsumerFactory;
+import org.springframework.util.ObjectUtils;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-
-import static java.util.Collections.emptyList;
 
 /**
  * Metrics for Kafka binder.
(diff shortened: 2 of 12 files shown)


mbogoevici:
Thanks @hekonsek,

This is a great contribution!

hekonsek:
@mbogoevici I'm glad you like it. :) 

This PR is a kind of proof of concept and I have not tested in a regular application yet. Wanted to check if you guys think this is something worth contributing. Tomorrow I will try to test it against live Kafka broker and in a standalone fat jar to be sure that everything works as I imagined. I will let you know when I'm done - I will also add documentation.

hekonsek:
Marius, I have polished the code and tested it in a standalone Spring Cloud Stream application. 

For standalone test I'm producing 100 messages every minute and slowing down consumer sink to consume one message per second. I'm invoking `/metrics` HTTP endpoint and I can see how input backlog is changing accordingly to those events. So now I can monitor my endpoints and tell Kubernetes to scale up/down my application if needed to deal with a significant backlog faster. :)

Can you take a look at the code and tell me if it is OK according to you? If you give me green light I will add documentation.

hekonsek:
Hi guys. Any chance to get review of what I've done so far? Sorry for being impatient... ;)

mbogoevici:
Hi Henryk,

We've been a little backed up (me in particular due to a holiday in the US). I'll get to a first round of review in the next couple of days. Thanks for the contribution (and the patience :))!

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -0,0 +1,137 @@
...
+				}
+				metrics.add(new Metric<>(String.format("%s.%s.%s.lag", METRIC_PREFIX, binding, group), lag));
+			} catch (Exception e) {
+				LOG.debug("Cannot generate metric for topic: " + topic, e);
+			}
+		}
+		return metrics;
+	}
+
+	private ConsumerFactory<?,?> createConsumerFactory(String group) {
```
mbogoevici: I wonder if this logic cannot be extracted as a utility class, since we seem to create these helper consumers in a few places already. Maybe additional refactoring after this PR, but this is an observation.

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderAutoConfigurationPropertiesTest.java
```diff
@@ -42,8 +36,15 @@
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 import org.springframework.util.ReflectionUtils;
 
+import java.lang.reflect.Field;
```
mbogoevici: Wrong import order - see https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style#import-statements - can be fixed on merge

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -0,0 +1,137 @@
...
+	@Override public Collection<Metric<?>> metrics() {
+		List<Metric<?>> metrics = new LinkedList<>();
+		for (Map.Entry<String, KafkaMessageChannelBinder.TopicInformation> topicInfo : this.binder.getTopicsInUse().entrySet()) {
+			if (!topicInfo.getValue().isConsumerTopic()) {
+				continue;
+			}
+
+			String topic = topicInfo.getKey();
+			String binding = topic;
+			for(Map.Entry<String, BindingProperties> bindingProperties : bindingServiceProperties.getBindings().entrySet()) {
```
mbogoevici: Instead of doing this, and relying on `BindingServiceProperties` - can we simplify this, and add the consumer group as a property to `TopicInformation` ?  Thus we would know which ones are anonymous, and also allows supporting the very unlikely but theoretically possible situation where we have multiple consumer groups on the same topic (for different bindings).

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
```
hekonsek: Noted!

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderJaasInitializerListenerTest.java
```diff
@@ -57,5 +60,10 @@ public void testConfigurationParsedCorrectlyWithKafkaClient() throws Exception {
 	@SpringBootApplication
 	public static class SimpleApplication {
 
+		@Bean
+		BindingServiceProperties bindingServiceProperties() {
```
hekonsek: `BindingServiceProperties` injected into the lag metric requires this bean.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
hekonsek: I was also thinking about this, but preferred to keep the code simpler now to make it easier to review and merge. Can we postpone this to the moment after we merge this PR? I can handle this, but I prefer little steps approach. But it is up to you to decide :) .

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderAutoConfigurationPropertiesTest.java
hekonsek: Noted! I will fix it.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
hekonsek: Let me see if this is possible. That would be a better approach indeed!

hekonsek:
Hi Marius,

Many thanks for review. I will check out if we can add group information to the `TopicInformation` class and push appropriate commit if it is possible. Thanks again!

hekonsek:
Hi Marius,

Indeed `TopicInformation` can and should carry group id of the consumer. I planned to handle anonymous consumers differently, but `TopicInformation#consumerGroup` solves everything. Silly me - I should figure it out in the first place :) . 

I have updated the PR including this change and all the others you suggested. Also after this change metric is stored under `spring.cloud.stream.binder.kafka.someTopic.someGroup.lag` key.

I will add some extra unit test coverage while waiting for an another review.

Cheers!

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderJaasInitializerListenerTest.java
hekonsek: I removed those mock bindingServiceProperties beans in https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/143/commits/4c6db2be73249550363d4f852667ceb298d76dff, as we don't rely on this bean anymore on the behalf of `TopicInformation#consumerGroup`.

mbogoevici:
@hekonsek Thanks for the updates. LGTM in general, I need to play a bit with it more before merging. Would it be possible to include a documentation section on the metric? Also, a quick question about the future of this feature. With this PR, we publish the aggregate lag of all partitions assigned to a specific consumer. Would you think it useful to provide additional lag information for each partition - this would allow to use this to identify partition skew as well? In that case, should we perhaps rename the property to 'aggregateLag' ? I was also thinking of suggesting a small change in the structure of the metric `spring.cloud.stream.binder.kafka.someGroup.someTopic.lag` (this would IMO allow to add partitions in the future ala `spring.cloud.stream.binder.kafka.someGroup.someTopic-0.lag`, `spring.cloud.stream.binder.kafka.someGroup.someTopic-1.lag` etc. WDYT ?

hekonsek:
Hi Marius,

Sure, take your time. I will give it extra end-to-end manual test as well. I will also commit documentation.

When it comes to the per-partition metrics. While my top priority would be to understand which topic has to be processed faster - that's why I'd like to receive summarized metrics in the first place. But I think that adding per-partition metrics in the future might be useful as well. So +1 for changing the naming convention to be partition-friendly.

hekonsek:
One question Marius - what would be the best place to add documentation for this metric? Something like section "14.5 Kafka metrics", after "14.4 Dead-Letter Topic Processing" [1]? 

[1] http://docs.spring.io/spring-cloud-stream/docs/Chelsea.SR2/reference/htmlsingle/index.html#kafka-dlq-processing

mbogoevici:
@hekonsek Sure, that would be great, thank you!

hekonsek:
Hi Marius. I've just added documentation for the metric. Let me know if there is anything else I can do to help with this PR.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -0,0 +1,118 @@
...
+				long lag = 0;
+				for (Map.Entry<TopicPartition, Long> endOffset : endOffsets.entrySet()) {
+					OffsetAndMetadata current = metadataConsumer.committed(endOffset.getKey());
+					if (current != null) {
+						lag += endOffset.getValue() - current.offset();
+					} else {
+						lag += endOffset.getValue();
+					}
+				}
+				metrics.add(new Metric<>(String.format("%s.%s.%s.lag", METRIC_PREFIX, topic, group), lag));
```
mbogoevici: If we keep with the decision in comments, (which is also how the docs have it: PREFIX.group.topic.lag ) then this should be changed to reflect that. Will do on merge.

mbogoevici:
Rebased, squashed and merged as https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/9ad04882c8c36e812432a1c3c2a5148a7821dc6e with minor polishing.

hekonsek:
Many thanks for all your help, Marius!

PR #171, opened 2017-07-17
Remove `resetOffsets` option

### LINKED ISSUE #170: Remove `resetOffsets` as a consumer option
Currently we don't support resetting offsets from within the application, but we have preserved this property since 1.0 (when using the simple consumer API). We should remove it, because the feature is not supported anymore.

See comments on #67 #63

mbogoevici:
Fix #170


ilayaperumalg:
LGTM, merging.

Issue #162, opened 2017-06-21
Drop Kafka 0.9 test module and support in 1.3


sobychacko:

mbogoevici:
Closed via https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/001cc5901e4b8d179237c8d12e13f28e2f011e2e

Issue #156, opened 2017-06-16
Drop Boot 1.4 support and simplify current code


mbogoevici:
Starting with version 1.3 of  Binder Kafka we can claim Spring Boot 1.5 support only, so we can ditch the conditional for backward compatibility with 1.4.

mbogoevici:
Closed via https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/001cc5901e4b8d179237c8d12e13f28e2f011e2e

Issue #129, opened 2017-05-03
KafkaBinderConfigurationProperties does not handle dot-named kafka configuration properties


joelmarty:
I recently updated spring-cloud-stream-binder-kafka to 1.2.0 from 1.1.2 using release train Camden.SR1 and somehow my application.yml that looked like this:

```yaml
spring.cloud.stream.kafka.binder:
  brokers: ${vcap.services.MH.credentials.kafka_brokers_sasl}
  configuration:
    security.protocol: SASL_SSL
    sasl.mechanism: PLAIN
    ssl.protocol: TLSv1.2
    ssl.enabled.protocols: TLSv1.2
    ssl.endpoint.identification.algorithm: HTTPS
```
did not work anymore. While debugging, I noticed that the field `configuration` in KafkaBinderConfigurationProperties did not contain the entries 'security.protocol' as expected but an entry 'security', containing a map with an entry 'protocol'. I had to change the configuration to the following to make it work:

```yaml
spring.cloud.stream.kafka.binder:
  brokers: ${vcap.services.MH.credentials.kafka_brokers_sasl}
  configuration:
    "[security.protocol]": SASL_SSL
    "[sasl.mechanism]": PLAIN
    "[ssl.protocol]": TLSv1.2
    "[ssl.enabled.protocols]": TLSv1.2
    "[ssl.endpoint.identification.algorithm]": HTTPS
```

Is it something that is expected ? I don't see how the version number change alone would have caused such behavior change.

I am using spring-boot-starter-parent 1.5.2.RELEASE.
I did not specified any specific spring cloud release train before but since my kafka service is based on version 0.10, I had to manually depend on spring-integration-kafka 2.1.0.RELEASE and kafka-clients 0.10.0.1, if this is in any way relevant.

ngawor:
I am experiencing the same with 1.3.1 (also using kafka client 0.10.0.1).

ngawor:
This is what my working application.properties ended up looking like:

spring.cloud.stream.kafka.binder.configuration.[security.protocol]=SASL_SSL
spring.cloud.stream.kafka.binder.configuration.[sasl.mechanism]=PLAIN
spring.cloud.stream.kafka.binder.configuration.[ssl.protocol]=TLSv1.2
spring.cloud.stream.kafka.binder.configuration.[ssl.enabled.protocols]=TLSv1.2
[...]

PR #172, opened 2017-07-17
Fixes for consumer and producer property propagation

### LINKED ISSUE #142: Problem with Serializers and Deserializers during channel binding
We're using SpringBoot/SpringCloudStream to stream information into Kafka.  There is an autoconfiguration in SpringBoot `org.springframework.boot.autoconfigure.kafka.KafkaProperties` that is setting both serializers and deserializers in the Producer and Consumer objects therein.

This is later referenced in `org.springframework.cloud.stream.binder.kafka.config. KafkaBinderConfiguration` in the PostConstruct because of the `@ConditionalOnClass(name = "org.springframework.boot.autoconfigure.kafka.KafkaProperties")`

As the `init()` method progresses, it constructs a `Map<String,Object> configuration` where all of the kafka configuration properties are dumped into one Map...  This is where the problem starts.

Later, when channels are bound, both the Producer and Consumer are fed serializers AND deserializers.  The `org.apache.kafka.clients.consumer.KafkaConsumer` and `org.apache.kafka.clients.producer.KafkaProducer` classes log the unnecessary configuration information at the WARN level (in the case of the Consumer, repeatedly on a scheduler)

I believe that an adjustment to the `init()` method in `KafkaPropertiesConfiguration` to filter out these fields (as you're only looking for binder information...), this will eliminate this misconfiguration.

Currently, I am able to circumvent the issue by setting:

- spring.kafka.consumer.key-deserializer=null
- spring.kafka.consumer.value-deserializer=null
- spring.kafka.producer.key-serializer=null
- sprint.kafka.producer.value-serializer=null

These are then not carried erroneously into the binder configuration.

If there were a filter put into the code
```
for (Map.Entry<String, Object> producerProperties : this.kafkaProperties.buildProducerProperties().entrySet()) {
	if (!configuration.containsKey(producerProperties.getKey())) {
		configuration.put(producerProperties.getKey(), producerProperties.getValue());
	}
}
for (Map.Entry<String, Object> consumerProperties : this.kafkaProperties.buildConsumerProperties().entrySet()) {
	if (!configuration.containsKey(consumerProperties.getKey())) {
		configuration.put(consumerProperties.getKey(), consumerProperties.getValue());
	}
}
```
such that it left out the properties if the key was in:

- ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG
- ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG
- ConsumerConfig.KEY_SERIALIZER_CLASS_CONFIG
- ConsumerConfig.VALUE_SERIALIZER_CLASS_CONFIG

mbogoevici:
Fix #142 #129 #156 #162

- Remove conditional configuration for Boot 1.4 support
- Filter properties before creating consumer and producer property sets
- Restore `configuration` as Map<String,String> for fixing Boot binding
- Remove 0.9 tests


garyrussell:
Merged as 001cc5901e4b8d179237c8d12e13f28e2f011e2e

PR #174, opened 2017-07-21
Porting KStream binder

### LINKED ISSUE #144: Move KStream binder to the main kafka binder repo
A fully working version of the kstream binder is here: https://github.com/mbogoevici/spring-cloud-stream-binder-kstream
This issue is for moving it to the main repo.

sobychacko:
Adding integration tests

Fix #144


spring-cloud-stream-binder-kstream/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+			<groupId>org.apache.kafka</groupId>
+			<artifactId>kafka-clients</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.kafka</groupId>
+			<artifactId>kafka-streams</artifactId>
+			<exclusions>
+				<exclusion>
+					<groupId>org.slf4j</groupId>
+					<artifactId>slf4j-log4j12</artifactId>
```
artembilan: Why do you duplicate `exclusion`?
You did already in the `dependencyManagement`.
Or doesn't it work as expected?

Thanks

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/KStreamDelegate.java
```diff
@@ -0,0 +1,245 @@
...
+import org.apache.kafka.streams.kstream.ValueJoiner;
+import org.apache.kafka.streams.kstream.ValueMapper;
+import org.apache.kafka.streams.kstream.ValueTransformerSupplier;
+import org.apache.kafka.streams.processor.ProcessorSupplier;
+import org.apache.kafka.streams.processor.StreamPartitioner;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class KStreamDelegate<K, V> implements KStream<K, V> {
```
artembilan: Should this really be `public`?
Where then JavaDocs?
I don't understand why do we need this class.

Can't we come up with something like `cache` on the `KStreamBoundElementFactory` to avoid delegation?
Because it seems form me we just propagate an `input` to the `output`, so there should be some correlation anyhow...

Thanks

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/config/KStreamBinderSupportAutoConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+@EnableKafkaStreams
+public class KStreamBinderSupportAutoConfiguration {
+
+	@Bean(name = KafkaStreamsDefaultConfiguration.DEFAULT_KSTREAM_BUILDER_BEAN_NAME)
+	public KStreamBuilderFactoryBean defaultKStreamBuilder(
+			@Qualifier(KafkaStreamsDefaultConfiguration.DEFAULT_STREAMS_CONFIG_BEAN_NAME) ObjectProvider<StreamsConfig> streamsConfigProvider) {
+		StreamsConfig streamsConfig = streamsConfigProvider.getIfAvailable();
+		if (streamsConfig != null) {
+			KStreamBuilderFactoryBean kStreamBuilderFactoryBean = new KStreamBuilderFactoryBean(streamsConfig);
+			kStreamBuilderFactoryBean.setPhase(Integer.MAX_VALUE - 500);
```
artembilan: What is the problem with this?
Maybe Spring Kafka should reconsider its `private int phase = Integer.MIN_VALUE;` then?

Otherwise you don't need `@EnableKafkaStreams` anyway, because you fully override it here.

spring-cloud-stream-binder-kstream/src/test/resources/logback.xml
```diff
@@ -0,0 +1,14 @@
...
+        </encoder>
+    </appender>
+    <logger name="org.springframework.integration.kafka" level="INFO"/>
+    <logger name="org.springframework.kafka" level="INFO"/>
+    <logger name="org.springframework.cloud.stream" level="INFO" />
+    <logger name="org.springframework.integration.channel" level="INFO" />
+    <root level="WARN">
+        <appender-ref ref="stdout"/>
+    </root>
+</configuration>
```
artembilan: New line in the end of file.

spring-cloud-stream-binder-kstream/pom.xml
mbogoevici: I think it's only copy-paste error when moving things together.

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/KStreamDelegate.java
mbogoevici: It doesn't have to be public. I'm actually thinking of how to get rid of the class (also because of the interface coupling), but the main problem that it's been solving was a disconnect between the binding lifecycle and the `@StreamListener` annotation processing. I'll provide more details after I look again at it.

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/config/KStreamBinderSupportAutoConfiguration.java
mbogoevici: @artembilan Yes, we might want to reconsider what Spring Kafka does. I'd like for SCSt to keep the annotation.

spring-cloud-stream-binder-kstream/src/test/resources/logback.xml
mbogoevici: We might want to get rid of this file, I think - at least the log format override.

mbogoevici:
@sobychacko @artembilan Added this commit with further refactoring: https://github.com/mbogoevici/spring-cloud-stream-binder-kafka/commit/ebe4b69d50a5fc5b3da1facb64a1c4c50b425a96

sobychacko:
In order for https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/174/commits/133ef17b53077db90218e122d7b9a80f499fcf82 in this PR, https://github.com/spring-cloud/spring-cloud-stream/pull/1031 needs to be reviewed/merged first.

spring-cloud-stream-binder-kstream/src/test/resources/logback.xml
artembilan: Hasn't been addressed.
Or have you decided to leave it here anyway?

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/KStreamBoundElementFactory.java
```diff
@@ -0,0 +1,168 @@
...
+import org.springframework.messaging.converter.MessageConverter;
+import org.springframework.messaging.support.MessageBuilder;
+import org.springframework.util.Assert;
+import org.springframework.util.MimeType;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class KStreamBoundElementFactory extends AbstractBindingTargetFactory<KStream> {
```
artembilan: I still think that we can get rid of row types...

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/config/KStreamExtendedBindingProperties.java
```diff
@@ -0,0 +1,61 @@
...
+	public Map<String, KStreamBindingProperties> getBindings() {
+		return this.bindings;
+	}
+
+	public void setBindings(Map<String, KStreamBindingProperties> bindings) {
+		this.bindings = bindings;
+	}
+
+	@Override
+	public KStreamConsumerProperties getExtendedConsumerProperties(String channelName) {
```
artembilan: I guess `channelName` is too strong already with all these binding abstractions.

spring-cloud-stream-binder-kstream/src/test/java/org/springframework/cloud/stream/binder/kstream/KStreamBinderPojoInputAndPrimitiveTypeOutputTests.java
```diff
@@ -0,0 +1,158 @@
...
+		assertThat(cr.key().equals(123));
+		assertThat(cr.value().equals(1L));
+	}
+
+	@EnableBinding(KStreamProcessor.class)
+	@EnableAutoConfiguration
+	public static class ProductCountApplication {
+
+		@StreamListener("input")
+		@SendTo("output")
```
artembilan: I wonder if we can make it send to the `output` by default, when we don't have this annotation or its `value` is empty.
The same is applied for the `@StreamListener` for its `target` if its empty.

WDYT?

Not related to this work thought. Just a thought until I remember it.

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/config/KStreamBinderSupportAutoConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	@Bean(KafkaStreamsDefaultConfiguration.DEFAULT_STREAMS_CONFIG_BEAN_NAME)
+	public StreamsConfig streamsConfig(KStreamBinderProperties kStreamBinderProperties) {
+		Properties props = new Properties();
+		props.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, kStreamBinderProperties.getKafkaConnectionString());
+		props.put(StreamsConfig.KEY_SERDE_CLASS_CONFIG, Serdes.ByteArraySerde.class.getName());
+		props.put(StreamsConfig.VALUE_SERDE_CLASS_CONFIG, Serdes.ByteArraySerde.class.getName());
+		props.put(StreamsConfig.APPLICATION_ID_CONFIG, "default");
```
mbogoevici: You can configure it via `spring.cloud.stream.kstream.binder.configuration.application.id`. I can make it a constant, but we need a value. Probably not 'default'.

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/config/KStreamBinderSupportAutoConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.integration.codec.Codec;
+import org.springframework.kafka.annotation.KafkaStreamsDefaultConfiguration;
+import org.springframework.kafka.core.KStreamBuilderFactoryBean;
+import org.springframework.util.ObjectUtils;
+
+/**
+ * @author Marius Bogoevici
+ */
+@EnableBinding
```
mbogoevici: Looks like a fluke.

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/config/KStreamBinderSupportAutoConfiguration.java
artembilan: `streamConfiguration` ?
`KStreamBinderProperties`.

I won't mind for constant, though.

mbogoevici: I was actually planning to rename it to `configuration`, just like all the others.

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/config/KStreamBinderSupportAutoConfiguration.java
mbogoevici: It's easy to reproduce with this PR. Just remove the bean definition and see the tests fail. The underlying problem is that the `OutputBindingLifecycle` (which does the bindings and redirects the output stream via `KStream.to()`) is in a later phase than one in which the bean factory starts. Looking back, it doesn't feel like there is any clear reason why `KStreamBuilderFactoryBean` should start as early as it does in Spring Kafka - but you won't notice that unless there's some work done in a `SmartLifecycle` bean. We can continue the discussion in Spring Kafka though. For now, I would leave it as is here (I removed the `@EnableKafkaStreams` annotation), until we have the fix in a Spring Kafka version that is managed by Spring Boot.

spring-cloud-stream-binder-kstream/src/test/resources/logback.xml
mbogoevici: Actually, I noticed this a test resource so we don't care all that much. I'll remove it - more log info is better for diagnostic anyway.

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/config/KStreamBinderSupportAutoConfiguration.java
artembilan: Good. I see. So, let's make that in Spring Kafka! The issue or simple PR would be great. You see I released Milestone recently but forgot to include this fix...

spring-cloud-stream-binder-kstream/src/test/resources/logback.xml
artembilan: Would be better to leave. We have logging config for tests in each SI module. My concern was here only about new line in the of file...

spring-cloud-stream-binder-kstream/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-binder-test</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-binder-kafka-test-support</artifactId>
+			<version>1.3.0.BUILD-SNAPSHOT</version>
+			<scope>test</scope>
+		</dependency>
```
mbogoevici: As a rule, we should not specify versions (especially project versions) in submodules too ...

mbogoevici: Also, this artifact does not exist anymore ...

spring-cloud-stream-binder-kstream/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-jar-plugin</artifactId>
+				<version>3.0.2</version>
+				<executions>
+					<execution>
+						<goals>
+							<goal>test-jar</goal>
```
mbogoevici: It's an execution not an exclusion. But we don't need to produce the test artifacts. I think this is an artifact of copy-paste.

artembilan: Doh. But glad to be useful at this point anyway 😄

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/KStreamBoundElementFactory.java
mbogoevici: This one is hard because of how the superclass is structured (the `super(KStream.class)` call in the constructor).

artembilan:
````java
super((Class<KStream<?, ?>>) (Class<?>) KStream.class);
````

😄

**Comment by mbogoevici** [comment]
@sobychacko An additional commit for fixing compile errors against latest SCSt SNAPSHOT (as of https://github.com/spring-cloud/spring-cloud-stream/commit/6f1bc59ee232e72f22aa635fc2c4c7bbaf0e19b9) : https://github.com/mbogoevici/spring-cloud-stream-binder-kafka/commit/074e97c0b2ad04274eb6064368a28097aca97458

File: spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/KStreamBoundElementFactory.java (hunk 5 above)
Comment (mbogoevici) [inline spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/KStreamBoundElementFactory.java]: For some reason, it looks worse :).

File: spring-cloud-stream-binder-kstream/pom.xml [hunk 12]
```diff
@@ -0,0 +1,84 @@
...
+	<description>Kafka Streams Binder Implementation</description>
+
+	<parent>
+		<groupId>org.springframework.cloud</groupId>
+		<artifactId>spring-cloud-stream-binder-kafka-parent</artifactId>
+		<version>1.3.0.BUILD-SNAPSHOT</version>
+	</parent>
+
+	<dependencies>
+		<dependency>
```
artembilan: Why all these dependencies are still here?
Have you had a chance to play with them somehow or at least answer why the concern hasn't been addressed?

I don't mind to address it in the separate PR, though, but some feedback would be great to have any way.

Thanks

spring-cloud-stream-binder-kstream/src/main/java/org/springframework/cloud/stream/binder/kstream/KStreamBinder.java
```diff
@@ -0,0 +1,188 @@
...
+					valueSerde = Utils.newInstance(properties.getExtension().getValueSerde(), Serde.class);
+					if (valueSerde instanceof Configurable) {
+						((Configurable) valueSerde).configure(streamsConfig.originals());
+					}
+				}
+			}
+			catch (ClassNotFoundException e) {
+				throw new IllegalStateException("Serde class not found: ", e);
+			}
+			outboundBindTarget.to((Serde<Object>) keySerde, (Serde<Object>) valueSerde, name);
```
artembilan: image


That's might not be critical, but I guess some warn in logs during build is present.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/7884d59bdc56b1d0b2378025277e0d5137b2b513

My congratulations!

Really cool work!

Yep! Looking forward for more, especially in the Kinesis Binder 😉

Issue #169, opened 2017-07-11
Partitioning should use the existing partition count


mbogoevici:
Instead, it seems to use the `partitionCount` property of the topic.

Originally it was: https://github.com/spring-cloud/spring-cloud-stream/issues/977

garyrussell:
This was broken long ago, perhaps when partitioning was moved from the binder to a channel interceptor in https://github.com/spring-cloud/spring-cloud-stream/commit/bd92af47840e861afccb3a1a263216311e400743#diff-290b09119bc6747e86a6d2720a0c52a7

So, it happens far-removed from the binder where knowledge of the physical partition count resides.

We might need to add a header to the message to tell the interceptor how many partitions exist.

cc/ @viniciusccarvalho @elnur

PR #184, opened 2017-08-18
GH-169: Use the Actual Partition Count (Producer)

garyrussell:
Fixes https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/169

If the configured `partitionCount` is less than the physical partition count on an existing
topic, the binder emits this message:

    The `partitionCount` of the producer for topic partJ.0 is 3, smaller than the actual partition count of 8 of the topic.
    The larger number will be used instead.

However, that is not true; the configured partition count is used.

Override the configured partition count with the actual partition count.


viniciusccarvalho:
Merged via 02913cd177ff540e64f591a54a6cb39318b0c660

PR #189, opened 2017-09-08
KStream binder: support class for application level properties

### LINKED ISSUE #188: Provide TimeWindow operation properties in KStream as first class binder properties
Currently, there is no easy way to configure TimeWindow properties such as `windowLength` and `advanceBy` through the binder. Users have to define and configure them in the applications. It will be convenient if these properties are provided out of the box in the Kstream binder properties so that end users don't have to define them in each application.

sobychacko:
Provide commonly used KStream application properties for convenient access at runtime

Fix #188


garyrussell:
Merged as 6485bd2abdcb9e663f684391dbdbaa0201ca8581

PR #191, opened 2017-09-09
Kstream binder: producer default Serde changes

### LINKED ISSUE #190: KStream binder: Default key or value Serde for producer
For the producer, when a user provides either key or value Serde, the Serde that is missing defaults to `ByteArraySerializer`. Before setting this, we need to check if there is a default common Serde set at the application level using `spring.cloud.stream.kstream.binder.configuration.key.serde` or `spring.cloud.stream.kstream.binder.configuration.value.serde`. 

Here is an example: A user provides key Serde for the producer, but not the value Serde which defaults to `ByteArraySerializer`. If the value is anything other than a byte array, then this will cause `ClassCaseException`.  However, if the user sets a common Serializer (Such as a `StringSerializer`) then we should give that a chance before falling back to `ByteArraySerializer`.

sobychacko:
Change the way the default Serde classes are selected for key and value
in producer when only one of those is provided by the user.

Fix #190


garyrussell:
Merged as 1aecd024041601ea05f3cb93dcd8c3f1340fefba

Issue #215, opened 2017-09-29
KafkaBinderHealthIndicator blocking application without timeout


fabianbuch:
KafkaBinderHealthIndicator is blocking and waiting for 5 minutes without timeout. That leads to /health not returning DOWN, but waiting forever (or load balancer timeout).

Testcase (KafkaBinderHealthIndicatorTest):

```java
	@Test(timeout = 1000)
	public void kafkaBinderDoesNotAnswer() {
		final List<PartitionInfo> partitions = partitions(new Node(-1, null, 0));
		topicsInUse.put(TEST_TOPIC, new KafkaMessageChannelBinder.TopicInformation("group", partitions));
		given(consumer.partitionsFor(TEST_TOPIC)).willAnswer(new Answer<Object>() {
			@Override
			public Object answer(InvocationOnMock invocation) throws Throwable {
				final int fiveMinutes = 1000 * 60 * 5;
				Thread.sleep(fiveMinutes);
				return partitions;
			}
		});
		Health health = indicator.health();
		assertThat(health.getStatus()).isEqualTo(Status.DOWN);
	}
```

PR #216, opened 2017-09-29
GH-215: Add timeout to health indicator

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/215

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
@@ -19,10 +19,12 @@ package org.springframework.cloud.stream.binder.kafka;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.kafka.clients.consumer.Consumer;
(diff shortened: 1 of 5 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
```diff
@@ -33,42 +38,89 @@
...
-				List<PartitionInfo> partitionInfos = metadataConsumer.partitionsFor(topic);
-				for (PartitionInfo partitionInfo : partitionInfos) {
-					if (this.binder.getTopicsInUse().get(topic).getPartitionInfos().contains(partitionInfo)
-							&& partitionInfo.leader()
-									.id() == -1) {
-						downMessages.add(partitionInfo.toString());
+		final CountDownLatch latch = new CountDownLatch(1);
+		final AtomicReference<Health> health = new AtomicReference<>();
+		ExecutorService exec = Executors.newSingleThreadExecutor();
+		exec.execute(new Runnable() {
```
artembilan: I believe that it works as you designed, but wouldn't `ExecutorService.submit()` be cleaner solution?
At least we won't need extra `CountDownLatch` variable.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
```diff
@@ -33,42 +40,87 @@
...
-		try (Consumer<?, ?> metadataConsumer = consumerFactory.createConsumer()) {
-			Set<String> downMessages = new HashSet<>();
-			for (String topic : this.binder.getTopicsInUse().keySet()) {
-				List<PartitionInfo> partitionInfos = metadataConsumer.partitionsFor(topic);
-				for (PartitionInfo partitionInfo : partitionInfos) {
-					if (this.binder.getTopicsInUse().get(topic).getPartitionInfos().contains(partitionInfo)
-							&& partitionInfo.leader()
-									.id() == -1) {
-						downMessages.add(partitionInfo.toString());
+		final AtomicReference<Health> health = new AtomicReference<>();
```
artembilan: But we don't need `AtomicReference` as well.
We can use `submit(Callable<T>)`  and return `Health` from there and `get()` it from the `Future`, therefore.

garyrussell: Doh

artembilan:
Cherry-picked to `1.2.x` as https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/657e8a2d7c54866250b7ee716b217705c0a61b9a

PR #227, opened 2017-10-11
Test structure improvements

### LINKED ISSUE #200: Refactor Kafka binder test-module
1) The reason why we have a separate module for Kafka tests is that we have a dependency to the Confluent maven repositories for schema registry testing. We did not want to introduce a 3rd part maven repository to the main binder pom. 

2) Consider moving the schema registry integration tests as samples to the samples repository, then removing the Confluent maven repositories from the pom. If we get there, then there is no need to have a separate test module. We can collapse this module (tests) into the main binder.

sobychacko:
Remove the separate test module introduced in 1.x to test different versions of Kafka.
In 2.0, there is only a single Kafka version that needs to be tested.
Move all the tests from the test module to the main binder module.
Remove the confluent schema registry integration test from the binder tests as it
will be ported as a sample application. This test currently does the serialziation/deserialition twice.

Fix #200


sobychacko:
Issue created in the samples repo for adding the schema registry sample: https://github.com/spring-cloud/spring-cloud-stream-samples/issues/46

olegz:
Merged

Issue #236, opened 2017-10-20
Investigate 0.10 and 0.11 broker compatibility with SCSt 1.3 and 2.0 apps


sabbyanandan:
As a developer, I'd like to study the broker and SCSt compatibilities. 

**Acceptance:**
The goal is to strive for backward and forward compatibility between the 0.10/0.11 brokers and SCSt versions.
- Study whether the 0.10 broker based logs are compatible with 0.11 broker
- Investigate SCSt 1.x and 2.0 app compatibility in the context of 0.10 and 0.11 brokers
- Document refactoring, testing, and migration-path efforts

PR #237, opened 2017-10-23
GH-236: Backwards Compatibility

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/236

- Restore the mapped headers configuration for embedding headers
- Configure the header mapper based on the header mode
- On the outbound side, set the mapper to `null` unless the header mode is not null and not `headers`
  this suppresses the channel adapter from setting up headers - for compatibility with < 0.11 brokers
- On the inbound side, set the `BinderHeaders.NATIVE_HEADERS_PRESENT` header if native headers found
- Add a configuration capability allowing the user to provide his/her own header mapper

- Add a test case with 3 producers (embedded, native, and no headers) and a consumer to verify it can
  consume all 3

Requires https://github.com/spring-cloud/spring-cloud-stream/pull/1107

Resolves spring-cloud/spring-cloud-stream#1106

### DIFF
--- a/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
+++ b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
@@ -27,6 +27,7 @@ import org.apache.kafka.clients.producer.ProducerConfig;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.kafka.KafkaProperties;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.kafka.support.DefaultKafkaHeaderMapper;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
@@ -97,6 +98,11 @@ public class KafkaBinderConfigurationProperties {
 	private int healthTimeout = 60;
 
 	private JaasLoginModuleConfiguration jaas;
+
+	/**
+	 * The bean name of a custom header mapper to use instead of a {@link DefaultKafkaHeaderMapper}.
+	 */
+	private String headerMapperBeanName;
 
 	public Transaction getTransaction() {
 		return this.transaction;
@@ -356,6 +362,14 @@ public class KafkaBinderConfigurationProperties {
 
 	public void setJaas(JaasLoginModuleConfiguration jaas) {
 		this.jaas = jaas;
+	}
+
+	public String getHeaderMapperBeanName() {
+		return this.headerMapperBeanName;
+	}
+
+	public void setHeaderMapperBeanName(String headerMapperBeanName) {
+		this.headerMapperBeanName = headerMapperBeanName;
 	}
 
 	public static class Transaction {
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
```diff
@@ -98,6 +99,11 @@
 
 	private JaasLoginModuleConfiguration jaas;
 
+	/**
+	 * The bean name of a custom header mapper to use instead of a {@link DefaultKafkaHeaderMapper}.
+	 */
+	private String headerMapperBeanName;
```
artembilan: Why can't we rely on the dependency injection?
I mean we provide the default `DefaultKafkaHeaderMapper` bean in the `KafkaBinderConfiguration` with the `@ConditionalOnMissingBean` and, therefore, if end-user provides its own bean on the matter, we will get exactly that injected.

At least that is how the `ProducerListener` is going to work. And that is how I do in the Kinesis Binder for the `MetadataStore`: https://github.com/spring-cloud/spring-cloud-stream-binder-aws-kinesis/blob/master/spring-cloud-stream-binder-kinesis/src/main/java/org/springframework/cloud/stream/binder/kinesis/config/KinesisBinderConfiguration.java#L83

WDYT?

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -194,19 +213,32 @@ protected MessageHandler createProducerMessageHandler(final ProducerDestination
...
-		if (headerPatterns != null && headerPatterns.length > 0) {
-			List<String> patterns = new LinkedList<>(Arrays.asList(headerPatterns));
-			if (!patterns.contains("!" + MessageHeaders.TIMESTAMP)) {
-				patterns.add(0, "!" + MessageHeaders.TIMESTAMP);
+		KafkaHeaderMapper mapper;
+		if (this.configurationProperties.getHeaderMapperBeanName() != null) {
+			mapper = getApplicationContext().getBean(this.configurationProperties.getHeaderMapperBeanName(),
+					KafkaHeaderMapper.class);
+		}
+		if (producerProperties.getHeaderMode() != null
```
artembilan: `else if()` ?

Because it looks now like we always override it here or in the next `else`

garyrussell: That was my intent - even if the user provides a bean, if the header mode is `raw` ,`none` or `embeddedHeaders` we need to signal to the channel adapter that it should not populate headers.

garyrussell: But I see the else needs help to not overwrite.

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
garyrussell: That won't work because `headerPatterns` can be specified at the binding level, so we need an instance per binding.

Also, the user might want a different mapper for each binding.

Issue #251, opened 2017-11-08
We need properties to configure the inbound MessagingMessageConverter


garyrussell:
The inbound channel adapter does not set ID and timestamp headers by default (for performance due to the cost of getting a UUID and timestamp).

However, certain SI components (such as aggregator - or any message store) require message ids.

Expose properties to configure the converter (and also allow a custom converter bean).

See [gitter chat](https://gitter.im/spring-projects/spring-integration?at=5a03221ce44c43700ad1fd70).

PR #255, opened 2017-11-10
GH-251: Configurable inbound adapter converter

garyrussell:
Fixes https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/251

When using native decoding, the message emitted by the binder has no `id` or `timestamp`
headers. This can cause problems if the first SI endpoint uses a JDBC message store,
for example.

Add the ability to configure whether these headers are populated as well as the ability
to inject a custom message converter.


garyrussell:
Cherry-pick to 1.3.x. 0.11.

garyrussell:
Rebased.

sobychacko:
Merged the changes to master. Backported to 1.3.x 0.11.

Issue #58, opened 2016-10-12
Add the ability to configure the DLQ producer properties


mbogoevici:
Identify what properties we want the users to set and propagate over to the DLQ configuraiton. (eg., we don't want to set serde)

ilayaperumalg:
Currently, the DLQ uses the default producer properties [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L287) and it would be better we provide support for all the Kafka producer properties for producing into DLQ topic rather than choosing specific properties.

aldex32:
Hi guys, any update about this issue? It's very important for people who are using `Confluent` serializer/deserializer.

Thanks in advance,
Aldo

sobychacko:
@aldex32 I want to understand your use case a bit further. We are definitely adding the way to configure dlq producer. However, I have one question about the usage of serde on dlq. The data that you are receiving on `dlq` is whats already coming from Kafka `ConsumerRecord` which is the `byte[]` representation. So, lets say that you used `nativeEncoding` on your producer, which means that Kafka used a serializer to serialize the data you provided. Then on the receiving side, some error happened, and the data is sent to `dlq`. This data is the same `byte[]` representation of what was originally encoded by the serializer natively in Kafka. Why do you want to re-serialize that same information? Or does it mean that you want to use a different `Serializer` at that point for `dlq`? Can you please clarify? Thanks!

garyrussell:
@sobychacko When using native (de)serialization, the inbound payload is not `byte[]`, it's whatever the deserializer created (probably a POJO). So, we need to be able to re-serialize that POJO back to a `byte[]` for kafka; hence the need to specify a value (and possibly key) serializer.

PR #258, opened 2017-11-14
Addressing PR review comment for gh-58

sobychacko:
- Add a check for ignoring dlq producer properties set by the user
  if transaction management is enabled
- Polishing

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -439,9 +439,10 @@ public class KafkaMessageChannelBinder extends
 				final ConsumerRecord<?, ?> record = message.getHeaders()
 						.get(KafkaHeaders.RAW_DATA, ConsumerRecord.class);
 
-				if (this.transactionManager == null && extendedConsumerProperties.isUseNativeDecoding()) {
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -420,26 +420,28 @@ protected ErrorMessageStrategy getErrorMessageStrategy() {
...
 
 			return message -> {
 				final ConsumerRecord<?, ?> record = message.getHeaders()
 						.get(KafkaHeaders.RAW_DATA, ConsumerRecord.class);
 
-				if (extendedConsumerProperties.isUseNativeDecoding()) {
+				if (this.transactionManager == null && extendedConsumerProperties.isUseNativeDecoding()) {
 					if (record != null) {
-						Map<String, String> configuration = extendedConsumerProperties.getExtension().getDlqProducerProperties().getConfiguration();
+						Map<String, String> configuration = dlqProducerProperties.getConfiguration();
```
garyrussell: I think you misunderstood - we still need to check that the global/transactional producer properties are compatible with the consumer.

Something like

```java
Map<String, String> configuration = this.transactionManager == null ? dlqProducerProperties.getConfiguration()
    : this.configurationProperties.getTransaction().getProducer().getConfiguration();
```

...and run `ensureDlqMessageCanBeProperlySerialized` against one or the other.

PR #264, opened 2017-11-17
Upgrade to Kafka 1.0.0

### LINKED ISSUE #259: Add Kafka 1.0 compatibility

sobychacko:
Resolves #259

- Remove the kafka server dependency from build/binder artifact
- Remove AdminUtilOperation and KafkaAdminUtilOperation
- Rely on AdminClient for provisioning operations
- Update KStream components with the new changes
- Test updates
- Polishing


spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
```diff
@@ -58,18 +69,23 @@
...
-	public KafkaTopicProvisioner(KafkaBinderConfigurationProperties kafkaBinderConfigurationProperties,
-								AdminUtilsOperation adminUtilsOperation) {
+	private int operationTimeout = DEFAULT_OPERATION_TIMEOUT;
+
+	public KafkaTopicProvisioner(KafkaBinderConfigurationProperties kafkaBinderConfigurationProperties) {
 		this.configurationProperties = kafkaBinderConfigurationProperties;
-		this.adminUtilsOperation = adminUtilsOperation;
+		Map<String, Object> adminConfigs = new HashMap<>();
+		adminConfigs.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaBinderConfigurationProperties.getKafkaConnectionString());
+		this.adminClient = AdminClient.create(adminConfigs);
```
garyrussell: We should use boot`s `KafkaProperties` bean and `buildAdminProperties()` - then override with any binder-specific properties such as the broker address.

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
```diff
@@ -127,30 +149,67 @@ public ConsumerDestination provisionConsumerDestination(final String name, final
...
+			if (throwable instanceof Error) {
+				throw (Error) throwable;
+			}
+			else {
+				throw new ProvisioningException("provisioning exception", throwable);
+			}
+		}
+	}
+
 	private void createTopicsIfAutoCreateEnabledAndAdminUtilsPresent(final String topicName, final int partitionCount,
```
garyrussell: Change this method name now we're not using `AdminUtils`.

spring-cloud-stream-binder-kafka-core/pom.xml
```diff
@@ -38,10 +38,6 @@
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka_2.11</artifactId>
-        </dependency>
         <dependency>
```
garyrussell: Nice!

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -1663,6 +1372,517 @@ public void testAnonymousGroup() throws Exception {
 		binding2.unbind();
 	}
 
+	@Test
+	@SuppressWarnings("unchecked")
```
garyrussell: I assume these test methods were moved for some reason; it makes it hard to review what changed, though.

sobychacko: There are two different types of tests in this suite, one that uses the `getBinder` from the `AbstractTestBinder` and another one with custom configuration properties. I simply moved the latter set to the bottom and cleaned up a few things.

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
garyrussell: We'll need to be careful to explain that boot's `failFast` property won't take effect (the binder will now always start even if there are failures provisioning - when https://github.com/spring-cloud/spring-cloud-stream/pull/1131 is applied.

garyrussell: (Or we could make it fail by throwing an `IllegalStateException` if there's a failure and the flag is true.

sobychacko: Addressed the first comment. Can't the second comment be addressed as a docs story? (or a separate PR for the 3rd comment).

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
```diff
@@ -58,18 +69,26 @@
...
 
+	private int operationTimeout = DEFAULT_OPERATION_TIMEOUT;
+
 	public KafkaTopicProvisioner(KafkaBinderConfigurationProperties kafkaBinderConfigurationProperties,
-								AdminUtilsOperation adminUtilsOperation) {
+								KafkaProperties kafkaProperties) {
+		Assert.isTrue(kafkaProperties != null, "KafkaProperties cannot be null");
+		Map<String, Object> stringObjectMap = kafkaProperties.buildAdminProperties();
+		//override if there is a binder level bootstrap server config or if it is different from base properties.
+		stringObjectMap.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaBinderConfigurationProperties.getKafkaConnectionString());
```
garyrussell: This defaults to `localhost:9092` - that is bad because it will override the boot property.

We need to implement a hierarchy - only use the binder stuff if the properties are actually set.

You might need to look at https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/6a31e9c94f604f7054dd4f50f905da4d8edd5c78 to see how @ilayaperumalg implemented it for the binder itself.

garyrussell: Maybe add `getDefaultKafkaConnectionString()` to the properties.

Then

```java
if (ObjectUtils.isEmpty(props.get(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG))
         || !this.configurationProperties.getKafkaConnectionString().equals(this.configurationProperties.getDefaultKafkaConnectionString())) {
	props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, this.configurationProperties.getKafkaConnectionString());
}

The only case that won't work for is if there is a boot property and they explicitly want to connect to `localhost:9092` - but that's rather unlikely.

File: pom.xml [hunk 6]
```diff
@@ -12,9 +12,9 @@
 	</parent>
 	<properties>
 		<java.version>1.8</java.version>
-		<spring-kafka.version>2.0.0.RELEASE</spring-kafka.version>
-		<kafka.version>0.11.0.0</kafka.version>
-		<spring-integration-kafka.version>3.0.0.M2</spring-integration-kafka.version>
+		<spring-kafka.version>2.1.0.RC1</spring-kafka.version>
```
garyrussell: Soby, please note that the 2.0.0.RELEASE renames the `KStreamBuilderFactoryBean` to align with renaming in the kafka 1.0.0 client: https://github.com/spring-projects/spring-kafka/pull/500

garyrussell:
Test failure for me...

```
Tests run: 55, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 67.567 sec <<< FAILURE!
testResume(org.springframework.cloud.stream.binder.kafka.KafkaBinderTests)  Time elapsed: 5.341 sec  <<< ERROR!
java.lang.IllegalArgumentException: Failed to receive message
	at org.springframework.util.Assert.isTrue(Assert.java:116)
	at org.springframework.cloud.stream.binder.kafka.KafkaBinderTests.testResume(KafkaBinderTests.java:1851)
```

sobychacko:
@garyrussell Could you run the build again? It passed for me locally. `testResume` is a test that fails randomly (on CI as well).

garyrussell:
@sobychacko I just ran a build against SIK 3.0.0.RELEASE (which is in the staging repo) and the tests pass.

WIll take one more pass in the morning and we can merge.

2.1.0 is released so could you fix up the KStream stuff?

garyrussell:
Merged as 50b8955dfc6330c150195d84a9c64c1ba95b0bef

PR #266, opened 2017-12-21
Fix Mixed Header Mode Tests

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/1159

- interceptor was missing


spring-cloud-stream-binder-kafka/.settings/org.eclipse.jdt.ui.prefs
```diff
@@ -0,0 +1,5 @@
+eclipse.preferences.version=1
```
artembilan: Why do we need two of this file, I see in each module?
Doesn't it work just on the root, project level?

Otherwise LGTM and I'll merge after your answer.

Thanks

garyrussell: 2? There are 3 - one for each project.

artembilan: Projects?

I thought we talk about modules in the `spring-cloud-stream-binder-kafka` project.

OK. If that how it works, let it be. 
Merging...

Issue #277, opened 2018-01-08
Add support for DeferredAcknowledgmentMessageSource


sabbyanandan:
As a continuation of spring-projects/spring-integration#2310, as a developer, I'd like to add support for `DeferredAcknowledgmentMessageSource` in the kafka-binder, so we could support pollable-consumer use-cases.

PR #282, opened 2018-01-16
GH-277: Add Polled Consumer

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/277

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2017 the original author or authors.
+ * Copyright 2014-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -387,6 +391,68 @@ public void stop(Runnable callback) {
...
+		kafkaMessageDrivenChannelAdapter.setBeanFactory(this.getBeanFactory());
+		ErrorInfrastructure errorInfrastructure = registerErrorInfrastructure(destination, consumerGroup,
+				extendedConsumerProperties);
+		if (extendedConsumerProperties.getMaxAttempts() > 1) {
+			kafkaMessageDrivenChannelAdapter.setRetryTemplate(buildRetryTemplate(extendedConsumerProperties));
+			kafkaMessageDrivenChannelAdapter.setRecoveryCallback(errorInfrastructure.getRecoverer());
+		}
+		else {
+			kafkaMessageDrivenChannelAdapter.setErrorChannel(errorInfrastructure.getErrorChannel());
+		}
```
olegz: Gary, can you please raise an "improvement" issue in core to revisit the above code? It is almost identical in Rabbit binder and in new SI-backed binder, so it seems to be begging for some abstraction and/or utility. So I'd like to revisit it at some point in the future.

garyrussell: The problem is that the abstract class doesn't know anything about the channel adapter and wouldn't know how to inject the error channel. This is related to the complete overhaul that Vinny wanted to do in the fall but found that it was a huge change. Perhaps this is something we can look at making the binders/bindings more spring-friendly.

Issue #305, opened 2018-02-13
Rework KMCB.ProducerConfigurationMessageHandler


garyrussell:
Recent changes to SIK to support gateways mean that `handleMessageInternal` is no longer overridable.

Fixes in SIK mean this is no longer necessary to support producer-initiated transactions.

PR #307, opened 2018-02-14
GH-305: Fix producer initiated transactions

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/305

Requires https://github.com/spring-projects/spring-integration-kafka/pull/196

`ProducerConfigurationMessageHandler` overrode `handleMessageInternal` to support
producer-initiated transactions. Now that the superclass extends `ARPMH`, this
method is no longer overridable.

Spring Integration Kafka's `KafkaProducerMessageHandler` now detects a transactional
`KafkaTemplate` and will automatically start a transaction if needed, so there is no
longer any need to override that method.

No additional tests needed; see `KafkaTransactionTests`.


artembilan:
This fixes #308 as well.

Issue #318, opened 2018-02-27
KafkaBinderMetrics are not exported with Spring Boot 2.0


urgelacko:
I am using Spring Cloud stream 2.0.0-RC1 with `micrometer-registry-graphite-1.0.0` and I have problems with exporting Kafka topic lag metrics.

1. Binder config is in separate context so the applications `MeterBinder` mechanism isn't using it. We can't even autowire it to build some adapter for it.
2. If you import the `KafkaBinderConfiguration` explicitly with `@Import` the `MeterBinder` is hooked up however when `bindTo` is called during application startup the `this.binder.getTopicsInUse()` returns 0 because the Kafka connections wasn't built up yet. Additionally it seems the explicit import results some beans are duplicated in the context.

olegz:
@urgelacko just to confirm. You are trying to `@Autowire` MeterBinder. And while successful it doesn't produce the expected results. Correct?

artembilan:
The `Micormeter` upgrade has been done as a part of the fix for the #193. Although it isn't obvious by the commit comment.

Just for info.

I think we really can't do import for the `KafkaBinderConfiguration` because the real Binder and all its goodies are definitely triggered in the separate application context as you say, @urgelacko. 
See `DefaultBinderFactory`:
```
SpringApplicationBuilder springApplicationBuilder = new SpringApplicationBuilder()
					.sources(binderType.getConfigurationClasses())
					.bannerMode(Mode.OFF)
					.logStartupInfo(false)
					.web(WebApplicationType.NONE);
```

The Spring Boot uses this approach in the `MeterRegistryPostProcessor`:
```
this.configurer = new MeterRegistryConfigurer(beansOfType(MeterBinder.class),
...
	private <T> Collection<T> beansOfType(Class<T> type) {
		return this.context.getBeansOfType(type).values();
	}
```

That's how our `KafkaBinderMetrics` isn't visible.

Not sure how to fix though, but at least this is a food for thoughts.

artembilan:
Well, we might need to do something similar what we have in the `BindersHealthIndicatorAutoConfiguration` with the `BindersHealthIndicatorListener` and its `afterBinderContextInitialized`. So, this way we can expose target `MeterBinder` beans from the binder context to this parent context and it will be available for the `MeterRegistryPostProcessor` in Spring Boot Actuator.

I may miss something anyway and we have some hook on the core level...

//CC @olegz , @sobychacko

[...]

PR #320, opened 2018-02-27
GH-318: Fix KafkaBinderMetrics for Micrometer

artembilan:
Fixes spring-cloud/spring-cloud-stream-binder-kafka#318

* Use `ToDoubleFunction`-based `MeterRegistry.gauge()` variant to really
calculate a `lag` at runtime instead of statically defined before
* Add `KafkaBinderActuatorTests` integration test to demonstrate how
`Binder` is declared in the separate child context and how
`KafkaBinderMetrics` is not visible from the parent context.
This test also verify the real `gauge` value for the `consumer lag`
and should be used in the future to verify the `KafkaBinderMetrics`
exposure removing the code after TODO in the `KafkaMetricsTestConfig`


artembilan:
Superseded by #321

PR #321, opened 2018-02-28
Pr/320

olegz:


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -53,72 +61,93 @@
...
 		}
-		if (!props.containsKey(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG)) {
-			props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG,
-					this.binderConfigurationProperties.getKafkaConnectionString());
+
+		return this.defaultConsumerFactory;
+	}
+	
+	@Override
+	public void onApplicationEvent(BindingCreatedEvent event) {
```
artembilan: Right, the event is here because you publish it from the Binder per se, into its own child context, not to the parent.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/353c89ab638161672c7b9695c29411c7ec2cde26 and https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/227d8f39f61de95165a48b0fa6e93f3a00ef2ffe

PR #325, opened 2018-03-02
Change KafkaBinderMetrics to a tagged TimeGauge

jkschneider:
Using a `TimeGauge` rather than a plain gauge ensures that the value is scaled to a base unit of time that matches what each monitoring system expects.

I'm a little distressed by the [implementation](https://github.com/omkreddy/kafka-examples/blob/master/consumer/src/main/java/kafka/examples/consumer/processor/AsyncRecordProcessor.java#L58-L64) in Kafka, because it appears to use `System#currentTimeMillis` to arrive at a millisecond latency. `currentTimeMillis` is _not_ a monotonically increasing clock. It can actually move back/forward on NTP shifts or local time discontinuities. But oh well.

Also moved group and topic to tags.

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
@@ -20,10 +20,13 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


jkschneider:
@artembilan Oops! `KafkaBinderActuatorTests` fixed.

olegz:
@artembilan while I agree to move `KafkaBinderMetrics` to a new class given that MR is optional, i would prefer it to be a separate PR/Issue for tracking purposes.

olegz:
@jkschneider you missing an `@author` tag. I'll add it during the merge (just wanted to let you know for the future)

Issue #326, opened 2018-03-03
KafkaProperties is not found when explicit binder configuration is used


sobychacko:
Without explicitly adding `KafkaAutoConfiguration` to the binder configuration, KafkaProperties is not found when using explicit (and multiple) binder configurations.

garyrussell:
This is because the provisioner now has a hard reference to `KafkaProperties` - I don't know why boot doesn't process the `KafkaAutoConfiguration` when we create the binder `SpringApplication`.

Also this: https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java#L48

(required = false)

is bogus; it means binders configured via `environment` properties won't inherit boot props.

garyrussell:
Looks like the problem is the child is not a `@SpringBootApplication`.

This fails too

```java
@SpringBootApplication
public class NestedApplication {

	public static void main(String[] args) {
		ConfigurableApplicationContext ctx = new SpringApplicationBuilder(NestedApplication.class)
				.run();
		KafkaProperties props = ctx.getBean(KafkaProperties.class);
		System.out.println("main: " + props.getBootstrapServers());
		ctx.close();
	}

	@Bean
	public ApplicationRunner runner() {
		return args -> {
			SpringApplicationBuilder builder = new SpringApplicationBuilder(com.example2.Nested.class);
			ConfigurableApplicationContext ctx = builder.run();
			System.out.println(ctx.getBean(String.class));
			try {
				KafkaProperties props = ctx.getBean(KafkaProperties.class);
				System.out.println("nested: " + props.getBootstrapServers());
			}
			catch (Exception e) {
				e.printStackTrace();
			}
			ctx.close();
		};
	}

}
```

and

```java
@Configuration
public class Nested {

	@Bean
	public String string() {
		return "foo";
	}

}
```

```
foo
org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.boot.autoconfigure.kafka.KafkaProperties' available
```

If I change it to 

```java
@SpringBootApplication
public class Nested {

	@Bean
	public String string() {
		return "foo";
	}

}
```

I get

```
Caused by: javax.management.InstanceAlreadyExistsException: org.springframework.boot:type=Admin,name=SpringApplication
[...]

PR #333, opened 2018-03-06
GH-326 added KafkaAutoConfiguration to KafkaBinderConfiguration

olegz:
- added KafkaAutoConfiguration to the @Import of KafkaBinderConfiguration
- removed 'optional' flag for KafkaProperties from KafkaBinderConfigurationProperties
- fixed KafkaBinderAutoConfigurationPropertiesTest

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java
@@ -26,6 +26,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.context.PropertyPlaceholderAutoConfiguration;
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java
```diff
@@ -55,7 +56,7 @@
  */
 @Configuration
 @ConditionalOnMissingBean(Binder.class)
-@Import({ PropertyPlaceholderAutoConfiguration.class, KafkaBinderHealthIndicatorConfiguration.class })
+@Import({KafkaAutoConfiguration.class, PropertyPlaceholderAutoConfiguration.class, KafkaBinderHealthIndicatorConfiguration.class })
```
artembilan: Is this a consequence of the fact that we configure a Binder in the child context?
Thanks

olegz: Yes, but most importantly this aligns it with the same way Rabbit binder is configured. See the actual issue for more details

artembilan: Right, and now I understand why we need it there and I know more about binders and their relationship with the parent application context.
Thanks
LGTM

PR #335, opened 2018-03-07
SCST-GH-1259: Kafka Binder Doc Polishing

garyrussell:
Fixes https://github.com/spring-cloud/spring-cloud-stream/issues/1259

Also deprecate properties that are no longer used.

### DIFF
--- a/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
+++ b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
@@ -110,6 +110,12 @@ public class KafkaBinderConfigurationProperties {
 		return this.transaction;
 	}
 
+	/**
+	 * No longer used.
+	 * @return the connection String
+	 * @deprecated connection to zookeeper is no longer necessary
+	 */
+	@Deprecated
 	public String getZkConnectionString() {
 		return toConnectionString(this.zkNodes, this.defaultZkPort);
 	}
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
```diff
@@ -110,6 +110,12 @@ public Transaction getTransaction() {
 		return this.transaction;
 	}
 
+	/**
+	 * No longer used.
+	 * @return the connection String
+	 * @deprecated connection to zookeeper is no longer necessary
+	 */
+	@Deprecated
```
artembilan: Looks like you deprecate here a configuration property.

Please, consider this additions: https://docs.spring.io/spring-boot/docs/2.0.0.RELEASE/reference/htmlsingle/#configuration-metadata-property-attributes

> Deprecation can also be specified declaratively in code by adding the `@DeprecatedConfigurationProperty` annotation ...

```
@DeprecatedConfigurationProperty(replacement = "app.acme.name")
	@Deprecated
	public String getTarget() {
```

I think this is how it will appear in the meta-data and be suggested via IDE auto-completion assistant.

garyrussell: As discussed on the standup - I think Oleg's intent is to just remove them during merge.

artembilan: Right, I don't mind. Just sharing the knowledge how properly deprecate configuration properties.

PR #346, opened 2018-03-27
Fix NPE in Kafka Streams binder

### LINKED ISSUE #343: Address NPE in kafka streams binder
if the users don't provide any producer/consumer properties, there is a [code path](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsStreamListenerSetupMethodOrchestrator.java#L294), that can throw an NPE.

sobychacko:
Fix NPE when user provided consumer properties are missing in kafka streams binder

Resolves #343

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsStreamListenerSetupMethodOrchestrator.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsStreamListenerSetupMethodOrchestrator.java
@@ -221,7 +221,7 @@ class KafkaStreamsStreamListenerSetupMethodOrchestrator implements StreamListene
 				enableNativeDecodingForKTableAlways(parameterType, bindingProperties);
 				StreamsConfig streamsConfig = null;
 				//Retrieve the StreamsConfig created for this method if available.
-				//Otherwise, carete the StreamsBuilderFactory and get the underlying config.
+				//Otherwise, create the StreamsBuilderFactory and get the underlying config.
 				if (!methodStreamsBuilderFactoryBeanMap.containsKey(method)) {
 					streamsConfig = buildStreamsBuilderAndRetrieveConfig(method, applicationContext, bindingProperties);
 				}
@@ -291,7 +291,10 @@ class KafkaStreamsStreamListenerSetupMethodOrchestrator implements StreamListene
 									Serde<?> keySerde, Serde<?> valueSerde) {
 		KStream<?, ?> stream = streamsBuilder.stream(bindingServiceProperties.getBindingDestination(inboundName),
 				Consumed.with(keySerde, valueSerde));
-		if (bindingProperties.getConsumer().isUseNativeDecoding()){
+		final boolean nativeDecoding = bindingProperties.getConsumer() != null &&
+				bindingProperties.getConsumer().isUseNativeDecoding();
+
+		if (nativeDecoding){
 			LOG.info("Native decoding is enabled for " + inboundName + ". Inbound deserialization done at the broker.");
 		}
 		else {
@@ -300,7 +303,7 @@ class KafkaStreamsStreamListenerSetupMethodOrchestrator implements StreamListene
 		stream = stream.map((key, value) -> {
 			KeyValue<Object, Object> keyValue;
 			String contentType = bindingProperties.getContentType();
-			if (!StringUtils.isEmpty(contentType) && !bindingProperties.getConsumer().isUseNativeDecoding()) {
+			if (!StringUtils.isEmpty(contentType) && !nativeDecoding) {
 				Message<?> message = MessageBuilder.withPayload(value)
 						.setHeader(MessageHeaders.CONTENT_TYPE, contentType).build();
 				keyValue = new KeyValue<>(key, message);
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsStreamListenerSetupMethodOrchestrator.java
```diff
@@ -291,7 +291,10 @@ else if (parameterType.isAssignableFrom(KTable.class)) {
 									Serde<?> keySerde, Serde<?> valueSerde) {
 		KStream<?, ?> stream = streamsBuilder.stream(bindingServiceProperties.getBindingDestination(inboundName),
 				Consumed.with(keySerde, valueSerde));
-		if (bindingProperties.getConsumer().isUseNativeDecoding()){
+		final boolean nativeDecoding = bindingProperties.getConsumer() != null &&
```
artembilan: Won't it be much safer to do this on the Spring Cloud Stream Core level in the `BindingProperties`:
```
private ConsumerProperties consumer = new  ConsumerProperties();

private ProducerProperties producer = new ProducerProperties();
```

At least this is what I do typically for any `@ConfigurationProperties`.

Also that looks strange that this object is `null` according Spring Boot Docs: https://docs.spring.io/spring-boot/docs/2.0.0.RELEASE/reference/htmlsingle/#boot-features-external-config-typesafe-configuration-properties

> If nested POJO properties are initialized (like the `Security` field in the preceding example), a setter is not required. If you want the binder to create the instance on the fly by using its default constructor, you need a setter.

So, or we use this class (`BindingProperties`) directly, then it is wrong and `consumer` and `producer`  have to be initialized on place, or something is wrong with Boot model so far.

On the other hand it is not clear why we have this API there in the `BindingServiceProperties`:
```
public BindingProperties getBindingProperties(String bindingName) {
	BindingProperties bindingProperties = new BindingProperties();
	if (this.bindings.containsKey(bindingName)) {
		BeanUtils.copyProperties(this.bindings.get(bindingName), bindingProperties);
	}
	if (bindingProperties.getDestination() == null) {
		bindingProperties.setDestination(bindingName);
	}
	return bindingProperties;
}
```

I know that there is no panacea from `NPE`, but I feel if we use conventions provided by Boot, we don't need to have such a protection.

I can merge, of course, as is if you are not agree with me (although... arguments to the studio!). Not a big deal. But you should be ready to meet similar problem in other places. 

BTW, why there is no test-case to demonstrate an issue? 
I'm asking because I am surprised to see this problem, when Spring Boot says that it is going to create one for us via its binder if we have setter.

sobychacko: Hi @artembilan if the user does not provide a `spring.cloud.stream.bindings.<input>.consumer` property, then boot won't create `ConsumerProperties` for us and that's why the NPE occurs. It's not required in Kafka streams binder and if that's the case, we just get by that.

sobychacko: the NPE can be reproduced by removing the conusmer properties from an existing test.

artembilan: OK! So, I think Boot can't create it for us because it is a nested in the `Map`:
```
private Map<String, BindingProperties> bindings = new TreeMap<>(
			String.CASE_INSENSITIVE_ORDER);
```
This is mostly question to Boot then...

Anyway: why don't use `new` then? And from there WDYT about my concern:
> But you should be ready to meet similar problem in other places.

My point is that I would prefer to see a fix in the Core, not here.
That's why I'm so picky.

And you if say that you know how to reproduce why don't cover the fix with the test?

sobychacko:
@artembilan i had that with `new` initially, but then there is no need for that as I can get by without creating that object. Basically, if the user doesn't provide a consumer property, there isn't really a need for that as we can infer the information through other means. But if you insist, I will switch to using `new`.  I will make a test.

artembilan:
Well, there must be some consistency though.

If we open, for example `KafkaExtendedBindingProperties`, it is similar to the mentioned `BindingServiceProperties`:
```
@ConfigurationProperties("spring.cloud.stream.kafka")
public class KafkaExtendedBindingProperties
		implements ExtendedBindingProperties<KafkaConsumerProperties, KafkaProducerProperties> {

private Map<String, KafkaBindingProperties> bindings = new HashMap<>();
```
Then when we jump into that `KafkaBindingProperties`, will see this:
```
private KafkaConsumerProperties consumer = new KafkaConsumerProperties();

private KafkaProducerProperties producer = new KafkaProducerProperties();
```
So, why here we have a `new`, but you insist that we should not in the `BindingServiceProperties`?

My apologies that I am so nasty, but I can't merge a fix without some ground of explanation and without understanding why do we have it from the first place at all.

Also an interesting observation: on the `BindingServiceProperties` we have this API:
```
public ConsumerProperties getConsumerProperties(String inputBindingName) {
```
Which protects us from this NPE as well:
```
ConsumerProperties consumerProperties = bindingProperties.getConsumer();
if (consumerProperties == null) {
	consumerProperties = new ConsumerProperties();
	bindingProperties.setConsumer(consumerProperties);
}
```

So, meanwhile we try to come up with the middle ground for the `new or not new` subject, I would suggest do not use `getConsumer()/getProducer()` API at all. 

My main point that I know it is about configuration properties, so Spring Boot will provide it for me this or other way without `null`. But that's not true in this case. Therefore for confuse with consistency.

sobychacko:
@artembilan I will use `BindingServiceProperties.getConsumerProperties(...)` as that seems to be a compromise here.

artembilan:
OK. I agree with that as state for the current code base.
Nevertheless are you going to raise an issue against Core to revise that `BindingProperties` for `consumer/producer` properties with `new` ?

Thanks

artembilan:
@sobychacko ,

I missed to ask: do we backport this into the `1.3.x` branch ?

sobychacko:
@artembilan No need to backport. Kafka streams binder has significantly changed in 2.0 and it doesn't make much sense to backport.

PR #362, opened 2018-04-14
Allow kafka-streams binder to consume messages from multiple kafka topics

### LINKED ISSUE #361: spring-cloud-stream-binder-kafka-streams does not support consuming from multiple kafka topics
The kafka-streams binder does not seem to work when multiple kafka topics are configured as input destinations.

From the spring cloud stream docs, 
> If the channel is bound as a consumer, it could be bound to multiple destinations, and the destination names can be specified as comma-separated String values. 

When multiple topic names are specified as comma separated strings, for instance, `spring.cloud.stream.bindings.input.destination=words1,words2` the entire string value is considered as the topic. Please refer `org.springframework.cloud.stream.binder.kafka.streams.KafkaStreamsStreamListenerSetupMethodOrchestrator#getkStream`

Unit test case attached
[KafkaStreamsBinderMultipleInputTopicsTest.java.zip](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/files/1910326/KafkaStreamsBinderMultipleInputTopicsTest.java.zip)

sarathshyam:
This is a proposal to fix the issue [#361](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/361). 
Changes

-  Modified `org.springframework.cloud.stream.binder.kafka.streams.KafkaStreamsStreamListenerSetupMethodOrchestrator#getkStream` so that the KStream object is built from list of topic names.
- Added new test case `org.springframework.cloud.stream.binder.kafka.streams.KafkaStreamsBinderMultipleInputTopicsTest`

Fixes #361


garyrussell:
While this might work fine for the Kafka Streams binder, I think there needs to be a more general solution for **all** binders that can support consuming from multiple topics (kafka, rabbitmq etc message channel binders, where the listener containers can be configured to consume from a lis of topics/queues).

olegz:
I agree with @garyrussell  and there is actually an issue that we already have for it https://github.com/spring-cloud/spring-cloud-stream/issues/1168

sarathshyam:
Thank you Soby

sobychacko:
@sarathshyam Thank you for the PR. This is merged upstream. A couple of notes related to this change. We introduced a new property at the core consumer level called ` multiplex`. When it is true, the binder natively is capable of consuming from multiple topics through a single input binding. This is always made true in the kafka streams binder. See the changes at this commit: https://github.com/spring-cloud/spring-cloud-stream/commit/61bce5272270b8f4e6b6ff8bed8c4be85450ae1d

Once this change was in the core framework, I added two more commits on top of your commit in the kafka streams binder. The following changes are introduced through those 2 commits.
* Spring Cloud Stream is updated to `2.1.0.BUILD-SNAPSHOT`
* `multiplex` property is made true always
*  When we have multiple inputs, the DLQ handling in kafka streams binder needs to be adjusted differently, so that we can have separate dlq topic per input topic or a catch all DLQ topic specified through the dlq name. These changes are also addressed through a commit in the PR. 

I did not squash the 2 commits that I added in the PR as they have individual focus but related to your base commit. 

Once again, thank you. Looking forward to more PR's in the future!

PR #379, opened 2018-05-08
Kafka Streams interactive query enhancements

### LINKED ISSUE #369: Interactive query issues when multiple Kafka Streams instances are used
When there are multiple instances of the same Kafka Streams application, support for interactive queries in the binder is unstable. Especially, when a particular state store is hosted on a different instance and the query is invoked against another instance, it may not produce correct results or throw exceptions. Binder should expose `StreamsMetadata` from the underlying `KafkaStreams` object so that the user can decide which host to invoke the query against.

sobychacko:
* When running interactive queries against multiple instances under the same application id,
  ensure that the application can retrieve the proper instance that is hosting the queried state store
* Introduce a new API level service called InteractiveQueryServices
* Perform refactoring to support this enhancement
* Deprecate QueryableStoreRegistry in 2.1.0 in favor of InteractiveQueryServices

Resolves #369


sobychacko:
Merging this PR.

PR #405, opened 2018-07-05
Upgrade kafka client to 1.1.0

### LINKED ISSUE #370: Upgrade Kafka Client to 1.1

sobychacko:
Upgrade kafka client to 1.1.0 for both kafka and kafka-streams binders

Resolves #370

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,9 +12,9 @@
 	</parent>
 	<properties>
 		<java.version>1.8</java.version>
-		<spring-kafka.version>2.1.5.RELEASE</spring-kafka.version>
+		<spring-kafka.version>2.1.7.RELEASE</spring-kafka.version>
 		<spring-integration-kafka.version>3.0.3.RELEASE</spring-integration-kafka.version>
-		<kafka.version>1.0.1</kafka.version>
+		<kafka.version>1.1.0</kafka.version>
 		<spring-cloud-stream.version>2.1.0.BUILD-SNAPSHOT</spring-cloud-stream.version>
 	</properties>
 	<modules>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -47,6 +47,12 @@
 				<artifactId>kafka-clients</artifactId>
 				<version>${kafka.version}</version>
 			</dependency>
+			<dependency>
+				<groupId>org.apache.kafka</groupId>
+				<artifactId>kafka-clients</artifactId>
+				<version>${kafka.version}</version>
+				<classifier>test</classifier>
+			</dependency>
```
garyrussell: I think this just needs to be in test scope; looks like a bug in our docs.

pom.xml
```diff
@@ -101,6 +107,27 @@
...
+					<exclusion>
+						<groupId>org.slf4j</groupId>
+						<artifactId>slf4j-log4j12</artifactId>
+					</exclusion>
+					<exclusion>
+						<groupId>log4j</groupId>
+						<artifactId>log4j</artifactId>
+					</exclusion>
+				</exclusions>
+			</dependency>
```
garyrussell: Why is this one in the parent but kafka itself is in the child poms?

pom.xml
sobychacko: will address this.

pom.xml
sobychacko: The ones in the parent pom are under the `dependencyManagment` section (Both kafka and the one with the `test` classifier). They both are in the relevant `dependencies` configuration in the child poms. Is that what you meant?

garyrussell: Yeah - I don't really know maven well enough to comment; it just looks a bit odd. I'll take your word if it's correct.

PR #407, opened 2018-07-05
Autoconfigure optimization

### LINKED ISSUE #406: Use Spring Boot autoconfigure-processor to optimize auto-configurations
As explained in [the its reference documentation](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc#autoconfigure-module), Spring Boot provides a `spring-boot-autoconfigure-processor`.

This should be added as an optional dependency to the module that contains the AutoConfiguration classes (and contains the `spring.factories` file declaring them).

Once configured, this annotation processor will generate metadata at compile time and will help Spring Boot process those auto-configurations without loading the actual classes in some cases.

The `spring-cloud-stream-binder-kafka-streams` is declaring at least one auto-configuration but doesn't use this annotation processor. Could you add it to the build?

```xml
<dependency>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-autoconfigure-processor</artifactId>
	<optional>true</optional>
</dependency>
```

Once added, this should generate a `spring-autoconfigure-metadata.properties` file in the `target/classes/META-INF` folder.

sobychacko:
Add spring-boot-autoconfigure-processor to the kafka-streams binder for
auto configuration optimization.

Resolves #406

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/pom.xml
+++ b/spring-cloud-stream-binder-kafka-streams/pom.xml
@@ -62,5 +62,10 @@
 			<artifactId>spring-cloud-stream-binder-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-autoconfigure-processor</artifactId>
+			<optional>true</optional>
+		</dependency>
 	</dependencies>
 </project>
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binder-kafka-streams/pom.xml
```diff
@@ -62,5 +62,10 @@
 			<artifactId>spring-cloud-stream-binder-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-autoconfigure-processor</artifactId>
```
artembilan: @sobychacko ,

don't we need to revise our all other binder implementation moving them to the normal Spring Boot auto-configuration model?

I somehow believe that the current `spring.binders` is fully similar what `spring.factories` provides. Therefore such an optimization should be done in all the binders.

BTW, shound't this fix be back-ported?

sobychacko: @artembilan, @olegz has already started on a change to use `spring.factories` instead of `spring.binders`, but preserved the usage of `spring.binders` for backward compatibility.

Yes, we need to backport this fix.  Thanks!

sobychacko:
Merged upstream.

PR #432, opened 2018-08-29
Handling tombstones in kafka streams binder

### LINKED ISSUE #294: KStreamListenerParameterAdapter crashes with NPE on Kafka tombstones records
_From @rliwoch on January 30, 2018 14:50_

Recently I've been working with a Kafka Connect connector (Debezium Mongo) which by default is utilising the Kafka topic compaction, it was not a desired behaviour for my use case so I've submitted a PR to their repository (worth checking as it will show how I got to the error I'm about to describe https://github.com/debezium/debezium/pull/421)  

The problem is that `KStreamListenerParameterAdapter` will not handle kafka `tombstones` records properly and it will crash with `NullPointerException` when it tries to consume a kafka message which is a `tombstone`.
```
		return bindingTarget.map(new KeyValueMapper() {
			@Override
			public Object apply(Object o, Object o2) {
				if (valueClass.isAssignableFrom(o2.getClass())) {
					return new KeyValue<>(o, o2);
				}
```
the `o2.getClass()` will generate `NPE`

Here is the snapshot from the topic, first message is a delete operation, the second one is the tombstone `null` associated (and generated) with this delete message.

And here's where the explosion is about to happen:

Detected in:
package: `org.springframework.cloud.stream.binder.kstream`
class: `KStreamListenerParameterAdapter`
maven co-ords to the jar: `org.springframework.cloud:spring-cloud-stream-binder-kstream11:1.3.0.RELEASE`
 

_Copied from original issue: spring-cloud/spring-cloud-stream#1199_

sobychacko:
Handle tombstones gracefully in the kafka streams binder
Modify tests to verify

Resolves #294

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsMessageConversionDelegate.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsMessageConversionDelegate.java
@@ -105,25 +105,24 @@ public class KafkaStreamsMessageConversionDelegate {
 				boolean isValidRecord = false;
 
 				try {
-					if (valueClass.isAssignableFrom(o2.getClass())) {
-						keyValueThreadLocal.set(new KeyValue<>(o, o2));
-					}
-					else if (o2 instanceof Message) {
-						if (valueClass.isAssignableFrom(((Message) o2).getPayload().getClass())) {
-							keyValueThreadLocal.set(new KeyValue<>(o, ((Message) o2).getPayload()));
+					//if the record is a tombstone, ignore and exit from processing further.
+					if (o2 != null) {
+						if (valueClass.isAssignableFrom(o2.getClass())) {
+							keyValueThreadLocal.set(new KeyValue<>(o, o2));
+						} else if (o2 instanceof Message) {
+							if (valueClass.isAssignableFrom(((Message) o2).getPayload().getClass())) {
+								keyValueThreadLocal.set(new KeyValue<>(o, ((Message) o2).getPayload()));
+							} else {
+								convertAndSetMessage(o, valueClass, messageConverter, (Message) o2);
+							}
+						} else if (o2 instanceof String || o2 instanceof byte[]) {
+							Message<?> message = MessageBuilder.withPayload(o2).build();
+							convertAndSetMessage(o, valueClass, messageConverter, message);
+						} else {
+							keyValueThreadLocal.set(new KeyValue<>(o, o2));
 						}
-						else {
-							convertAndSetMessage(o, valueClass, messageConverter, (Message) o2);
-						}
+						isValidRecord = true;
 					}
-					else if (o2 instanceof String || o2 instanceof byte[]) {
-						Message<?> message = MessageBuilder.withPayload(o2).build();
-						convertAndSetMessage(o, valueClass, messageConverter, message);
-					}
-					else {
-						keyValueThreadLocal.set(new KeyValue<>(o, o2));
-					}
-					isValidRecord = true;
 				}
 				catch (Exception ignored) {
 					//pass through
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/integration/KafkaStreamsBinderWordCountIntegrationTests.java
```diff
@@ -133,6 +138,18 @@ private void receiveAndValidate(ConfigurableApplicationContext context) throws E
...
+	private void sendTombStoneRecordsAndVerifyGracefulHandling() {
+		Map<String, Object> senderProps = KafkaTestUtils.producerProps(embeddedKafka);
+		DefaultKafkaProducerFactory<Integer, String> pf = new DefaultKafkaProducerFactory<>(senderProps);
+		KafkaTemplate<Integer, String> template = new KafkaTemplate<>(pf, true);
+		template.setDefaultTopic("words");
+		template.sendDefault(null);
+		ConsumerRecords<String, String> received = consumer.poll(Duration.ofMillis(5000));
+		//By asserting that the received record is empty, we are ensuring that the tombstone record
+		//was handled by the binder gracefully.
+		assertThat(received.isEmpty()).isTrue();
```
artembilan: You have to destroy `DefaultKafkaProducerFactory` in the end of class do not pollute a memory with non-used active objects.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsMessageConversionDelegate.java
```diff
@@ -105,25 +105,24 @@ public KStream deserializeOnInbound(Class<?> valueClass, KStream<?, ?> bindingTa
...
-						keyValueThreadLocal.set(new KeyValue<>(o, o2));
-					}
-					else if (o2 instanceof Message) {
-						if (valueClass.isAssignableFrom(((Message) o2).getPayload().getClass())) {
-							keyValueThreadLocal.set(new KeyValue<>(o, ((Message) o2).getPayload()));
-						}
-						else {
-							convertAndSetMessage(o, valueClass, messageConverter, (Message) o2);
+					//if the record is a tombstone, ignore and exit from processing further.
+					if (o2 != null) {
```
artembilan: Don't we need to do anything in the else?
Is logic in the end of the method going to work if we don't add anything into the `ThreadLocal` holder?

The `Object objectValue = keyValueThreadLocal.get().value;` is going to fail with `NPE`. No?

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsMessageConversionDelegate.java
```diff
@@ -164,22 +163,22 @@ public void init(ProcessorContext context) {
...
-				if (kstreamBindingInformationCatalogue.isDlqEnabled(bindingTarget)) {
-					String destination = context.topic();
-					if (o2 instanceof Message) {
-						Message message = (Message) o2;
-						sendToDlqAndContinue.sendToDlq(destination, (byte[]) o, (byte[]) message.getPayload(), context.partition());
-					}
-					else {
-						sendToDlqAndContinue.sendToDlq(destination, (byte[]) o, (byte[]) o2, context.partition());
+				//Only continue if the record was not a tombstone.
+				if (o2 != null) {
```
artembilan: Why can't we send to DLQ a null? Sounds like a valid use-case for Kafka.
Or at least some representation for null, or simple String mention that null is sent to DQL...

I will accept any explanation, but still would like to raise a concern for further thoughts.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsMessageConversionDelegate.java
sobychacko: The last block in the method will only be invoked if there is a stream coming through that branch. The first filter above that returns `false`, thus there is no stream coming through `branch[0]` and therefore it won't be invoked. So, I think we are safe there.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsMessageConversionDelegate.java
sobychacko: The record is a tombstone and I don't think we should send that to the DLQ as it is. I like the suggestion of a string representation of the tombstone record. Haven't seen such use cases or feature requests before, but worth looking into as a future enhancement.

artembilan: Right. That's fine to move it into the future. Meanwhile let's consider at least some INFO as an `else` variant. 
Otherwise there might be some question where has that gone...

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsMessageConversionDelegate.java
artembilan: OK. So, I need to understand first of all what is that `branch()` for. My home work though 😄 

Meanwhile maybe it make sense at least log INFO why have we stopped processing here. 
Does it make sense?

sobychacko: `else` case is added with logging.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsMessageConversionDelegate.java
sobychacko: The `else` case is addressed where the tombstone record is processed. This one is specifically for handling deserialization errors and since we already log the message as part of the initial processing, I don't think we need to log it again.

Issue #411, opened 2018-07-17
KStreamBinderConfiguration cannot find KafkaProperties when using multi-binder configuration


2speed:
**Environment**
- JDK - 1.8.0_172
- Gradle - 4.8.1
- Spring Boot - 2.0.3.RELEASE
- Spring Cloud Stream - Elmhurst.SR1

**Description**

I ran into an issue similar to spring-cloud/spring-cloud-stream-binder-kafka#326 using Spring Boot with the Kafka streams binder when attempting to bridge a stream from an external Kafka cluster. When running my application, I get the following error:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

Field kafkaProperties in org.springframework.cloud.stream.binder.kafka.streams.KStreamBinderConfiguration required a bean of type 'org.springframework.boot.autoconfigure.kafka.KafkaProperties' that could not be found.


Action:

Consider defining a bean of type 'org.springframework.boot.autoconfigure.kafka.KafkaProperties' in your configuration.
```

**Code**

`build.gradle`
```gradle
buildscript {
	repositories {
        mavenCentral()
        jcenter()
        maven { url 'https://repo.spring.io/plugins-release' }
        maven { url 'https://plugins.gradle.org/m2' }
        mavenLocal()
	}
	dependencies {
		classpath 'org.springframework.boot:spring-boot-gradle-plugin:2.0.3.RELEASE'
	}
}

apply plugin: 'java'
apply plugin: 'eclipse'
apply plugin: 'org.springframework.boot'
apply plugin: 'io.spring.dependency-management'

group = 'com.esm.stream'
version = '0.0.1-SNAPSHOT'
sourceCompatibility = 1.8

repositories {
	mavenCentral()
	jcenter()
}

ext {
	springCloudVersion = 'Elmhurst.SR1'
}

dependencies {
    // Spring : Test
    testCompile 'org.springframework.boot:spring-boot-starter-test'
    testCompile 'org.springframework.cloud:spring-cloud-stream-test-support'
    
    // Spring : SCS
    compile 'org.springframework.cloud:spring-cloud-stream-binder-kafka-streams'
}

dependencyManagement {
	imports {
		mavenBom "org.springframework.cloud:spring-cloud-stream-dependencies:${springCloudVersion}"
	}
}
```

`application.yml`
```yaml
spring.cloud.stream:
  bindings:
    input:
      binder: kBind1
  binders:
    kBind1:
[...]

PR #441, opened 2018-09-07
Binding support for GlobalKTable

### LINKED ISSUE #384: Add binding support for GlobalKTable

sobychacko:
* New binding targets and binder implementation for GlobalKTable within kafka-streams binder
* Refactoring existing structure to accommodate the new binder
* Adding integration test to verify the GlobalKTable behavior

Resolves #384


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBoundElementFactory.java
```diff
@@ -0,0 +1,92 @@
...
+
+	GlobalKTableBoundElementFactory(BindingServiceProperties bindingServiceProperties) {
+		super(GlobalKTable.class);
+		this.bindingServiceProperties = bindingServiceProperties;
+	}
+
+	@Override
+	public GlobalKTable createInput(String name) {
+		ConsumerProperties consumerProperties = this.bindingServiceProperties.getConsumerProperties(name);
+		//Always set multiplex to true in the kafka streams binder
```
artembilan: Is there some explanation "Why?" ? And from here: why isn't it by default in the Kafka Streams per se since it is so important for us?

Thanks

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsConsumerBindingUtils.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.cloud.stream.binder.kafka.streams.properties.KafkaStreamsConsumerProperties;
+import org.springframework.context.ApplicationContext;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Soby Chacko
+ */
+class KafkaStreamsConsumerBindingUtils {
+
+	public static void prepareConsumerBinding(String name, String group, Object inputTarget,
```
artembilan: Since class is not `public`, there is no value to make its methods `public`

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinder.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * {@link org.springframework.cloud.stream.binder.Binder} implementation for {@link GlobalKTable}.
+ * This implementation extends from the {@link AbstractBinder} directly.
+ *
+ * Provides only consumer binding for the bound GlobalKTable as output bindings are not allowed on it.
+ *
+ * @author Soby Chacko
+ */
+public class GlobalKTableBinder extends
+		AbstractBinder<GlobalKTable<Object, Object>, ExtendedConsumerProperties<KafkaStreamsConsumerProperties>, ExtendedProducerProperties<KafkaStreamsProducerProperties>>
```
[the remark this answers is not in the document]
sobychacko: maybe we can provide this as an enhancement later?

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @author Soby Chacko
+ */
+@Configuration
+public class GlobalKTableBinderConfiguration {
+
+	@Bean
+	@ConditionalOnBean(name = "outerContext")
+	public BeanFactoryPostProcessor outerContextBeanFactoryPostProcessor() {
```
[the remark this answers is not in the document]
sobychacko: Good idea about using the `ImportBeanDefinitionRegistrar`. Issue created to track that (https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/445). 

This bean registration was originally added as part of [this issue](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/411) in the binder and [this one](https://github.com/spring-cloud/spring-cloud-stream/issues/1420) in the core. We could revisit this as part of the issue created ^^.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBoundElementFactory.java
sobychacko: Since it is allowed to consume from multiple topics in a single binding, multiplex is enabled by default in the Kafka streams binder. In the regular Kafka binder, the user has to explicitly enable it. See this [javadoc](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ConsumerProperties.java#L144). As you suggested, It definitely makes sense to enable it by default across the various binding types in Kafka streams binder (`KStream`, `KTable` and `GlobalKTable`) instead of doing it individually.

artembilan: That's not what I'm asking about, but let it be.
I believe if this comment would be much informative, I'd not ask this question.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinder.java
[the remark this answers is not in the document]
artembilan: Huh? Why later?
This is a feature of this new Binder and what we have so far is confusing.
What is the point to commit wrong code and even release it when we already know that it is wrong and we see a solution.

What is the point of review then if we do nothing? 🤷‍♂️ 

Sorry, if I'm so nasty, but I don't think this is the way Spring evolve...

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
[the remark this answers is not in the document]
artembilan: Why revise later, but not in this PR? Why then we need review?
I agree that we can revise some other code, but this one is new. So, what is the point to still keep it wrong way, when we have a potential solution which could be then used for other wrong places if any?

Thanks for understanding.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinder.java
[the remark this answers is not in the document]
sobychacko: We throw `UnsupportedOperationException` from the producer binding method. So, anyone who is trying to bind an output will get that exception anyway. I don't see the necessity for doing that here. Maybe I am missing something?

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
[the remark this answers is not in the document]
sobychacko: Which one are you suggesting to do in this PR - `ImportBeanDefinitionRegistrar` or the refactoring of the parent-child relationship? I think we can do the former, but if its the latter, it might be better to address that as part of a separate issue as that involves some semantic changes.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinder.java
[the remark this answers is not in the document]
artembilan: Right, but we don't throw it from the `KafkaStreamsProducerProperties`. So, I will spend some time (a lot of time in case of YAML!) to configure producer properties just to fail in the end.
If we would have some blocker just from the API level, there is a little chance that people would try to configure producer properties.
Also some special no-op type wouldn't let us to fail with our internal logic and we wouldn't have any implementation for producer behavior.

Anyway I might just don't see the whole picture and all this discussion may just lead to dead end just because we may not achieve the solution with Java (or YAML?) restrictions.

So, let it really be and let's try to revise it later!

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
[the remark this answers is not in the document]
artembilan: Sorry, for confusion. Right, i mean that parent-child relationship is good, but that's really going to be much wider issue.
Right now I would just like to see an `ImportBeanDefinitionRegistrar` no any `beanFactory.getBean()` in the `BeanFactoryPostProcessor`.
You can simple check for the `outerContext` `BeanDefinition` before populating a `BeanDefinition` for your purpose.

The `beanFactory.getBean()` is very dangerous in this case. See "The early BeanFactory access syndrome": https://www.google.com/search?q=The+early+BeanFactory+access+syndrome&oq=The+early+BeanFactory+access+syndrome 😉

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
```diff
@@ -16,32 +16,54 @@
...
-		return beanFactory -> {
-			ApplicationContext outerContext = (ApplicationContext) beanFactory.getBean("outerContext");
-			beanFactory.registerSingleton(KafkaStreamsBinderConfigurationProperties.class.getSimpleName(), outerContext
-					.getBean(KafkaStreamsBinderConfigurationProperties.class));
-			beanFactory.registerSingleton(KafkaStreamsBindingInformationCatalogue.class.getSimpleName(), outerContext
-					.getBean(KafkaStreamsBindingInformationCatalogue.class));
-		};
+	static class Registrar implements ImportBeanDefinitionRegistrar {
+
+		private static final String BEAN_NAME = "outerContext";
```
artembilan: Looks like this constant is out of use.

spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/integration/StreamToGlobalKTableJoinIntegrationTests.java
```diff
@@ -214,7 +213,10 @@ public void testStreamToGlobalKTable() throws Exception {
 				assertThat(enrichedOrder.getCustomer().name.equals("customer-" + i)).isTrue();
 				assertThat(enrichedOrder.getProduct().name.equals("product-" + i)).isTrue();
 			}
+			pfCustomer.destroy();
+			consumer.close();
```
artembilan: There are `pfProduct`, `pfOrder` also to destroy.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
sobychacko: Actually we use this [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/454ca83d15ef5492e1db82eff7623b0f2f8aee49/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java#L48).

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+
+		@Override
+		public void registerBeanDefinitions(AnnotationMetadata importingClassMetadata,
+											BeanDefinitionRegistry registry) {
+			if (registry.containsBeanDefinition(BEAN_NAME)) {
+
+				AbstractBeanDefinition configBean = BeanDefinitionBuilder.genericBeanDefinition(MethodInvokingFactoryBean.class)
+						.addPropertyReference("targetObject", "outerContext")
```
artembilan: My point is that you don't use that `BEAN_NAME` constant here and for the next bean definition, too.

sobychacko: My bad. Will address that soon.

iamdaniu:
I'm still getting "No qualifying bean of type 'org.springframework.boot.autoconfigure.kafka.KafkaProperties' available" (cfg [Issue 411](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/411)) with a binder of type `globalktable` and two separate config files (to be able switch security on/off):

`application.yaml`:
```
spring.cloud.stream:
  bindings.cache:
    destination: items-topic
    binder: cache
    group: cache-group
  kafka.streams:
    binder:
      application-id: splitter
      brokers: localhost:9092
      configuration.default:
        key.serde: org.apache.kafka.common.serialization.Serdes$StringSerde
        value.serde: com.example.minfail.SampleSerdes$ItemSerde
    bindings:
      cache.consumer:
        applicationId: cache
        materializedAs: items-cache
  binders.cache.type: globalktable
```

`application-kerberos.yaml`:
```
spring.cloud.stream:
  kafka.binder.configuration:
    security.protocol: SASL_SSL
    sasl.kerberos.service.name: kafka
    jaas:
      loginModule: com.sun.security.auth.module.Krb5LoginModule
      options:
        useKeytab: true
        storeKey: true
  binders.cache.environment.spring.cloud.stream.kafka.binder.configuration.jaas.options:
    keyTab: cache_keytab
    principal: cache_principal
```
Relevant code:
```
@EnableBinding(CacheBinding.class)
public class ItemCacher {
    @StreamListener
    void materialize(@Input(CacheBinding.ITEMS) GlobalKTable<String, Item> items) {}
}

interface CacheBinding {
    String ITEMS = "cache";
    @Input(ITEMS) GlobalKTable<String, Item> cache();
}

@Data class Item {
    private String key;
    private String value;
}

@Data class ItemList {
    private List
<Item> items;
}
```

Above fails to start when activating the kerberos profile.

PR #450, opened 2018-09-20
Application ID resolution for Kafka streams binder

### LINKED ISSUE #448: Kafka Streams binder- application id checking precedence order
When checking for application id - the binder currently gives the highest precedence to `group` set at the consumer binding level.  This has to be restructured as below.

If there is an application ID set at the individual binding level, use that. For example,  `spring.cloud.stream.kafka.streams.bindings.input.conusumer.applicationId`. If there are multiple bindings within a single processor (`StreamListener`), then the applications need to ensure that the same `applicationId` is used for all the bindings. 

If there is only a single processor (`StreamListener`) within the app, then in order to avoid repetition, it should be able to configure `applicationId` through a default configuration. For instance, `spring.cloud.stream.kafka.streams.default.consumer.applicionId`. 

Binder currently allows to set `applicationId` using a global binder level property (`spring.cloud.stream.kafka.streams.binder.applicationId`). This may not be needed because of the default mechanism mentioned above. However, in order to preserve backward compatibility, we may want to keep it as it is. 

If the user doesn't provide an `applicationId` through any of the mechanisms mentioned above, the binder uses a default one, but keep in mind that this will have unexpected behavior if there are multiple `StreamListeners` in the application. 

With these changes, we can avoid relying on `group` property on consumer bindings to infer `applicaitonId`. The `group` configuration can still be used, where it is needed per core spring cloud stream requirements, but kafka streams binder will not require it from applications.

sobychacko:
Avoid the need to rely on group property for application id.
First, check binding specific application id, if not look at defaults.
If nothing works, fall back to the default application id set by the
boot auto configuration.

Modify tests.
Update docs.

Resolves #448

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
@@ -22,7 +22,6 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.stream.Collectors;
 
-import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.serialization.Serdes;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.errors.LogAndContinueExceptionHandler;
@@ -78,7 +77,7 @@ public class KafkaStreamsBinderSupportAutoConfiguration {
 		// Override Spring Boot bootstrap server setting if left to default with the value
 		// configured in the binder
 		if (ObjectUtils.isEmpty(properties.get(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG))) {
-			properties.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, binderConfigurationProperties.getKafkaConnectionString());
+			properties.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, binderConfigurationProperties.getKafkaConnectionString());
 		}
 		else {
 			Object bootstrapServerConfig = properties.get(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG);
@@ -87,7 +86,7 @@ public class KafkaStreamsBinderSupportAutoConfiguration {
 				String bootStrapServers = (String) properties
 						.get(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG);
 				if (bootStrapServers.equals("localhost:9092")) {
-					properties.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, binderConfigurationProperties.getKafkaConnectionString());
+					properties.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, binderConfigurationProperties.getKafkaConnectionString());
 				}
 			}
 		}
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
@@ -17,7 +17,6 @@
 package org.springframework.cloud.stream.binder.kafka.streams;
 
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.stream.Collectors;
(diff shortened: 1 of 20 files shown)


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
```diff
@@ -60,29 +64,61 @@ public KafkaStreamsBinderConfigurationProperties binderConfigurationProperties(K
...
+		if (StringUtils.isEmpty(streamsConfiguration.get(StreamsConfig.APPLICATION_ID_CONFIG))) {
+			streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, binderProvidedApplicationId);
+		}
+		else {
+			String applicationId = (String)streamsConfiguration.get(StreamsConfig.APPLICATION_ID_CONFIG);
+			if (applicationId.equals(environment.getProperty("spring.application.name")) &&
+					!StringUtils.isEmpty(binderProvidedApplicationId)) {
+				streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, binderProvidedApplicationId);
+			}
+		}
```
garyrussell: Shouldn't a binder-specific `applicationId` ALWAYS take precedence over the boot one, regardless of boot's value?

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
```diff
@@ -60,29 +64,61 @@ public KafkaStreamsBinderConfigurationProperties binderConfigurationProperties(K
...
-		props.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, binderConfigurationProperties.getKafkaConnectionString());
-		props.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.ByteArraySerde.class.getName());
-		props.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.ByteArraySerde.class.getName());
-		props.put(StreamsConfig.APPLICATION_ID_CONFIG, binderConfigurationProperties.getApplicationId());
+	public Map<String, Object> streamConfigGlobalProperties(KafkaStreamsBinderConfigurationProperties binderConfigurationProperties,
+															Environment environment) {
+
+		//Initialize configuration with default boot auto configuration properties
+		Map<String, Object> streamsConfiguration =
+				new HashMap<>(binderConfigurationProperties.getKafkaProperties().buildStreamsProperties());
```
garyrussell: I think you should use the `KafkaStreamsConfiguration` `@Bean` to take advantage of boot's auto-configuring the `spring.application.name` to the `application.id` if not explicitly set.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsStreamListenerSetupMethodOrchestrator.java
```diff
@@ -411,14 +418,22 @@ private void enableNativeDecodingForKTableAlways(Class<?> parameterType, Binding
...
+
+		KafkaStreamsConsumerProperties extendedConsumerProperties = kafkaStreamsExtendedBindingProperties.getExtendedConsumerProperties(inboundName);
+		//Need to apply the default extended properties here as it is not yet done by the BindingService in the binding lifecycle.
+		handleExtendedDefaultProperties(kafkaStreamsExtendedBindingProperties,
+				extendedConsumerProperties);
+
+		String applicationId = extendedConsumerProperties.getApplicationId();
+
+		//override application.id if set at the individual binding level.
+		if (!StringUtils.isEmpty(applicationId)) {
```
garyrussell: `hasText()` is better (trims).

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
```diff
@@ -59,30 +63,58 @@ public KafkaStreamsBinderConfigurationProperties binderConfigurationProperties(K
 		return new KafkaStreamsBinderConfigurationProperties(kafkaProperties);
 	}
 
+	@Bean
+	public KafkaStreamsConfiguration kafkaStreamsConfiguration(KafkaStreamsBinderConfigurationProperties binderConfigurationProperties) {
+		return new KafkaStreamsConfiguration(
+				new HashMap<>(binderConfigurationProperties.getKafkaProperties().buildStreamsProperties()));
+	}
```
garyrussell: Why are you overriding Boot's `@Bean` ? It has logic to use the `spring.application.name` if there is no ID.

Issue #455, opened 2018-09-26
How to Handle payload=org.springframework.kafka.support.KafkaNull?


adershrp:
Hi,
While we are using @StreamListener, is there any way to handle Null (KafkaNull) Payload?
I tried to use @Payload(required = false) in arguments, that is not working. I guess, this work well with @KafkaListener.

Another option I tried is enableDlq for the consumer, and this message goes in there with Null Payload. But, this message is of no use to me.

Please suggest, is there any better approach?

Regards,
Adersh

olegz:
No, there can never be a `Message` with _null_ payload. That won'y be allowed by the underlying framework (i.e., Spring Integration). Now, `KafkaNull` is not null. It's an object that ". . .represents NULL Kafka payload. . ." . This means in theory you should be able to receive `Message<KafkaNull>`. 
That said. . . .
Could you elaborate on your use case. I am not clear as to the value of null payload in general. 
Also, please use [Stack Overflow](https://stackoverflow.com/tags/spring-cloud-stream) channel to ask questions. We use GitHub for reporting issues only.

garyrussell:
Sorry for the delay; we were at SpringOne Platform last week.

Kafka records can have a null payload - this is termed a tombstone record for compacted logs (indicating the key has been deleted). Since a Spring-Messaging Message<?> can't have a `null` payload, we represent this condition with `KafkaNull.INSTANCE`.

In the `@KafkaListener` we have a customized payload resolver, which allows `(required=false)` on the payload...

```java
			argumentResolvers.add(new PayloadArgumentResolver(messageConverter) {

				@Override
				protected boolean isEmptyPayload(Object payload) {
					return payload == null || payload instanceof KafkaNull;
				}

			});
```

We should add similar logic to `@StreamListener`.

The outbound channel adapter already converts `KafkaNull` to `null` when passing to Kafka.

garyrussell:
Actually; since SCSt uses Spring Integration's `HandlerMethodArgumentResolversHolder`, the fix could be done there so SI apps can also benefit from it.

We'll just need a doc change here.

PR #458, opened 2018-10-04
GH-455: Docs for tombstone records

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/455


garyrussell:
(Requires latest Spring Integration 5.1.0.BUILD-SNAPSHOT).

PR #462, opened 2018-10-10
Setting group in TopicInformation correctly for anonymous consumers.

amanzag:
I'm not sure if the current behaviour is intended or not, but I just realised that metrics aren't collected for anonymous consumers because TopicInformation doesn't contain the group name.
This PR fixes it.

I couldn't find any tests verifying that TopicInformation is built correctly. If there is any, please point me to them.

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -380,12 +380,12 @@ public class KafkaMessageChannelBinder extends
 				+ "use the regex pattern to specify multiple topics instead");
 		boolean groupManagement = extendedConsumerProperties.getExtension().isAutoRebalanceEnabled();
 		if (!extendedConsumerProperties.isMultiplex()) {
-			listenedPartitions.addAll(processTopic(group, extendedConsumerProperties, consumerFactory,
+			listenedPartitions.addAll(processTopic(consumerGroup, extendedConsumerProperties, consumerFactory,
 					partitionCount, usingPatterns, groupManagement, destination.getName()));
 		}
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@amanzag Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream-binder-kafka&pullRequestId=462)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-stream-binder-kafka&pullRequestId=462) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@amanzag Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream-binder-kafka&pullRequestId=462)!

sobychacko:
@amanzag Thanks for the PR. Maybe you can create a `SpringBootTest` and then `autowire` the binder bean (`KafkaMessageChannelBinder`) and call `getTopicsInUse` to verify? (The test needs to be in the same package as the binder - `org.springframework.cloud.stream.binder.kafka`). At the very least, you can modify [this test](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/bootstrap/KafkaBinderBootstrapTest.java#L37) and query the binder bean directly from the application context and verify. One issue there though is that this test is not in the same package as the binder and thus you cannot invoke the `getTopicsInUse` method as it has default access. Anyway, these are some pointers to think.

amanzag:
Hi @sobychacko, I've added a couple of assertions to KafkaBinderTests, let me know what you think.
Thanks

sobychacko:
@amanzag Thank you for the contribution! The PR is now merged upstream.

PR #463, opened 2018-10-10
Kafka Health Indicator changes

### LINKED ISSUE #454: Unable to turn off kafka health indicator
Currently Health indicator is created on:
`@ConditionalOnClass(name="org.springframework.boot.actuate.health.HealthIndicator")`

And we can not simple disable it anyhow,

Can you add standard actuator mechanism?

`@ConditionalOnEnabledHealthIndicator("kafka")
@AutoConfigureBefore(HealthIndicatorAutoConfiguration.class)`

Then we could disable healthIndicator in actutor standard way by:
`managment.health.kafka.enabled=true`

sobychacko:
Disable Kafka health indicator check if management.health.binders.enabled
property is set to false. Currently, if this property is disabled, only the
core spring-cloud-stream mechanism of collecting the health checks are disabled.
The indivividual binders can still register the health check bean and boot health
actuator will still pick it up. If the user turns off health check by disabling
this property, then the stream app should completely disable any binder specific
health check.

Resolves #454


ilayaperumalg:
LGTM

PR #472, opened 2018-10-16
Support headers in Kafka Streams binder

### LINKED ISSUE #456: ContentType header value inconsistent between Kafka and Kafka Streams binders
Hi,
When using Avro for serialization, the spring-cloud-stream-binder-kafka and spring-cloud-stream-schema, the content type of the outbound message header reflects the avro type and version as documented e.g. application/vnd.user.v2+avro.

When using the spring-cloud-stream-binder-kafka-streams binder, the content type reflects the content type property for the output binding rather than the actual type resolved by the MessageConverter.

The conversion appears to be done here : [KafkaStreamsMessageConversionDelegate](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/e869516e3d0577db98d9c81e34e5e94e6200b7c7/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsMessageConversionDelegate.java#L76)

Perhaps a Processor could be added to update the header info using the Processor Context with the updated content type.

Spring Cloud : Greenwich.M1
Spring Cloud Stream : 2.1.0.M3
Spring Cloud Stream Binder Kafka Streams : 2.1.0.M3
Spring Cloud Stream Schema : 2.1.0.M3

Cheers.

### LINKED ISSUE #469: Kafka Streams and Spring Cloud Stream Schema Registry
With the new header support in Kafka Streams, we can now use Spring Cloud Stream provided Schema Registry with the Kafka Streams binder (#456). However, one that is addressed, the downstream Kafka Streams topology still needs to use an `Avro Serde` that can talk to the Spring Cloud Stream Schema Registry. This means that we need to expose our internal schema registry based `Avro` converter wrapped inside a `Serde`.

sobychacko:
* Use content type header from the record for binder provided inbound deserialization by
  making use of the header support added in Kafka Streams. If there is a content type set
  on the incoming record, that will get precedence.
* Introduce a new Composite Serde class for providing non-native Spring Cloud Stream specific collection
  of Serde implemenations. This is needed in order for things like avro converters that interact with
  the Spring Cloud Stream schema registry server.
* Adding tests
* Polishing

Resolves #456 
Resolves #469

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,6 +129,12 @@
 					</exclusion>
 				</exclusions>
 			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-stream-schema</artifactId>
+				<version>${spring-cloud-stream.version}</version>
+				<scope>test</scope>
+			</dependency>
 		</dependencies>
 	</dependencyManagement>
 
--- a/spring-cloud-stream-binder-kafka-streams/pom.xml
+++ b/spring-cloud-stream-binder-kafka-streams/pom.xml
@@ -13,6 +13,10 @@
 		<version>2.1.0.BUILD-SNAPSHOT</version>
 	</parent>
 
+	<properties>
+		<avro.version>1.8.2</avro.version>
+	</properties>
+
 	<dependencies>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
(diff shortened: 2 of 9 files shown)


olegz:
Resolved with https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/8ab289a715e44c3821e183b1ad2f8dcfc0e31215

olegz:
Resolved with https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/8ab289a715e44c3821e183b1ad2f8dcfc0e31215

olegz:
Resolved with https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/8ab289a715e44c3821e183b1ad2f8dcfc0e31215

Issue #376, opened 2018-05-05
Could not reset offsets to earliest to get the stream from beginning for KTable


srujanakuntumalla:
Hey,

I am using spring cloud kafka binder to read the data to KStream. While reading the data from one of the topic, i need to read from beginning.

I have tried to set kafka offset reset and start offset properties. But, could not find any references.

Could you please help me providing any sample application.yaml to reset the offset, so that i can consume messages from topic from the beginning

Adding application.yaml that i have used:

```
spring.cloud.stream.bindings.input:
  destination: input-topic1
  consumer:
    useNativeDecoding: true
    headerMode: raw
spring.cloud.stream.bindings.output:
  destination: output-topic
  producer:
    useNativeDecoding: true
    headerMode: raw
spring.cloud.stream.bindings.beginningInput:
  destination: beginning-topic
  consumer:
    useNativeDecoding: true
    headerMode: raw
spring.cloud.stream.kafka.streams.bindings.input:
  consumer:
    keySerde: org.apache.kafka.common.serialization.Serdes$StringSerde
    valueSerde: org.apache.kafka.common.serialization.Serdes$StringSerde
spring.cloud.stream.kafka.streams.bindings.output:
  producer:
    keySerde: org.apache.kafka.common.serialization.Serdes$StringSerde
    valueSerde: org.apache.kafka.common.serialization.Serdes$StringSerde
spring.cloud.stream.kafka.streams.bindings.beginningInput:
  consumer:
    keySerde: org.apache.kafka.common.serialization.Serdes$StringSerde
    valueSerde: org.apache.kafka.common.serialization.Serdes$StringSerde
    resetOffsets: true
    startOffset: earliest
spring.cloud.stream.kafka.streams.binder:
  brokers: 127.0.0.1
  zkNodes: 127.0.0.1
  configuration:
    default.key.serde: org.apache.kafka.common.serialization.Serdes$StringSerde
    default.value.serde: org.apache.kafka.common.serialization.Serdes$StringSerde
    commit.interval.ms: 1000
```

I would like to know if there is any error in the configuration i have used

PedroAlvarado:
Per the code below, it seems that it is not possible to set `resetOffsets` via kafka stream binder consumer properties.

[...]

PR #473, opened 2018-10-17
Addressing startOffset in Kafka Streams binder

### LINKED ISSUE #467: startOffset not honored by KStream/KTable creation
With `auto.offset.reset` (globally) set to “latest” and `startOffset` set to "earliest" for a specific KTable/KSteam:

```
spring.cloud.stream.kafka.streams.binder.auto.offset.reset: latest

spring.cloud.stream.kafka.streams.bindings.configInput:
  consumer:
    materializedAs: configState
    keySerde: org.apache.kafka.common.serialization.Serdes$StringSerde
    valueSerde: org.apache.kafka.common.serialization.Serdes$StringSerde
    startOffset: earliest
```

when the table/stream is created, only the global setting (“latest”) seems to take effect. i.e. The very first time this application comes up, only the latest offsets are processed.

I was able to confirm that `extendedConsumerProperties` is correctly populated with startOffset=“earliest” [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsStreamListenerSetupMethodOrchestrator.java#L246). However, it does not seem to be propagated down to the table/stream creation.

If I change [this](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsStreamListenerSetupMethodOrchestrator.java#L308) to include a `Consumed` object with a hard-coded “earliest”  then the KTable does behave as expected and consume from the earliest offset on the first start of the application. 

```
        private <K,V> KTable<K,V> materializedAs(StreamsBuilder streamsBuilder, String destination, String storeName, Serde<K> k, Serde<V> v) {
                return streamsBuilder.table(bindingServiceProperties.getBindingDestination(destination),
                                Consumed.with(k, v).withOffsetResetPolicy(Topology.AutoOffsetReset.EARLIEST),
                                Materialized.<K, V, KeyValueStore<Bytes, byte[]>>as(storeName)
                                                .withKeySerde(k)
                                                .withValueSerde(v));
        }
```

Is this expected? 

Thanks,
Alex

sobychacko:
Fixing the issue where auto.offset.reset is not honored through startOffset
provided by the individual consumer binding.

Adding integration test to override auto.offset.reset globally at the binder
level to latest and then set startOffset on individual binding to earliest.

Resolves #467

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/integration/StreamToTableJoinIntegrationTests.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/integration/StreamToTableJoinIntegrationTests.java
@@ -36,8 +36,6 @@ import org.apache.kafka.streams.kstream.Joined;
 import org.apache.kafka.streams.kstream.KStream;
 import org.apache.kafka.streams.kstream.KTable;
 import org.apache.kafka.streams.kstream.Serialized;
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/integration/StreamToTableJoinIntegrationTests.java
```diff
@@ -208,6 +197,142 @@ public void testStreamToTable() {
...
+		try (ConfigurableApplicationContext ignored = app.run("--server.port=0",
+				"--spring.jmx.enabled=false",
+				"--spring.cloud.stream.bindings.input.destination=user-clicks-2",
+				"--spring.cloud.stream.bindings.input-x.destination=user-regions-2",
+				"--spring.cloud.stream.bindings.output.destination=output-topic-2",
+				"--spring.cloud.stream.bindings.input.consumer.useNativeDecoding=true",
+				"--spring.cloud.stream.bindings.input-x.consumer.useNativeDecoding=true",
+				"--spring.cloud.stream.bindings.output.producer.useNativeEncoding=true",
+				"--spring.cloud.stream.kafka.streams.binder.configuration.auto.offset.reset=latest",
+				"--spring.cloud.stream.kafka.streams.bindings.input.consumer.resetOffsets=true",
```
garyrussell: I don't see this property being used anywhere in the streams binder; the message channel binder seeks to beginning/end when this is true.

sobychacko: Removed it. This was a leftover and an oversight from my side to not cleanup.

garyrussell: Hehe - my question is really "should it be honored?", or "should we detect it and reject the configuration since we don't honor it" 😄

sobychacko: Honoring `resetOffset` in the streams binder is going to be tricky. See [this discussion](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/376#issuecomment-388207759). This is because of all the intermediate changelog topics kafka streams create behind the scenes. So I think it is like a no-op configuration. If the user provides it, that has no effect. What do you mean by "detect and reject"? In effect, thats the end result now although the binder doesn't do any detection of that property.

garyrussell: Well, I don't think you would want to mess with the intermediate topics, just the input topic. But, I agree, it's messy if the intermediates aren't reset to the end when the input is reset (to either end).

It's your call, but throwing an exception if it's configured, but not used, might avoid user confusion. Or, at least add a note to the docs.

PR #475, opened 2018-10-19
Consumer concurrency settings in Kafka Streams binder

### LINKED ISSUE #474: Consumer Concurrency property has no effect
I configured consumer as per this [https://docs.spring.io/spring-cloud-stream/docs/Brooklyn.RELEASE/reference/html/_configuration_options.html](https://docs.spring.io/spring-cloud-stream/docs/Brooklyn.RELEASE/reference/html/_configuration_options.html)

```
spring-cloud-dependencies:Finchley.SR1
springBootVersion = '2.0.6.RELEASE'
```

There are 4 partitions for my topic and messages are filled in round robin by the producer.
```
root@kafka:/# kafka-run-class kafka.tools.GetOffsetShell --broker-list localhost:9092 --topic kstream_test --time -1
kstream_test:2:222
kstream_test:1:203
kstream_test:3:188
kstream_test:0:278
```
But the consumer concurrency property has no effect.

```
spring.cloud.stream.bindings.input:
  destination: kstream_test
  group: consumer-group-G1_test
  consumer:
    useNativeDecoding: true
    headerMode: raw
    startOffset: latest
    partitioned: true
    concurrency: 3
```
It still runs on 1 thread and not 3. The current work around as per my [SO question](https://stackoverflow.com/questions/52878308/spring-cloud-stream-kstream-consumer-concurrency-has-no-effect/52878849#52878849) is to use the below config.

```
spring.cloud.stream.kafka.streams.binder.configuration:
  num.stream.threads: 3
```

Sample project to reproduce the problem https://github.com/titoc/kstream-demo

sobychacko:
* If the consumer concurency settings are provided at the binding level,
  honor that before falling back to the defaults.
* Allow consumer binding specific broker configurations to be set from the application.
* Test changes.

Resolves #474

### DIFF
--- a/spring-cloud-stream-binder-kafka-docs/src/main/asciidoc/kafka-streams.adoc
+++ b/spring-cloud-stream-binder-kafka-docs/src/main/asciidoc/kafka-streams.adoc
@@ -84,7 +84,6 @@ For common configuration options and properties pertaining to binder, refer to t
 === Kafka Streams Properties
 
 The following properties are available at the binder level and must be prefixed with `spring.cloud.stream.kafka.streams.binder.`
-literal.
 
 configuration::
   Map with a key/value pair containing properties pertaining to Apache Kafka Streams API.
@@ -120,7 +119,7 @@ applicationId::
 +
 Default: `none`
 
-The following properties are _only_ available for Kafka Streams producers and must be prefixed with `spring.cloud.stream.kafka.streams.bindings.<binding name>.producer.` literal.
+The following properties are _only_ available for Kafka Streams producers and must be prefixed with `spring.cloud.stream.kafka.streams.bindings.<binding name>.producer.`
 For convenience, if there multiple output bindings and they all require a common value, that can be configured by using the prefix `spring.cloud.stream.kafka.streams.default.producer.`.
 
 keySerde::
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka-docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -173,6 +173,11 @@ Default: `earliest`.
 Note: Using `resetOffsets` on the consumer does not have any effect on Kafka Streams binder.
 Unlike the message channel based binder, Kafka Streams binder does not seek to beginning or end on demand.
 
+concurrency::
+  Number of stream task threads. Needs at least these many number of partitions in the consuming topic.
++
+Default: `1`.
+
```
garyrussell: Why is this here? `concurrency` is a common binding property, not a kafka (streams)-spedific property: [Common Binding Properties](https://docs.spring.io/spring-cloud-stream/docs/Elmhurst.SR1/reference/htmlsingle/#_common_binding_properties).

garyrussell: It certainly doesn't belong in this section, which talks about `spring.cloud.stream.kafka.streams.bindings.<binding name>.consumer.'literal`.

Perhaps add a note somewhere that the common concurrency property maps to the `num.stream.threads` kafka property.

BTW, there's something wrong with the `spring.cloud.stream.kafka.streams.bindings.<binding name>.consumer.'literal.` above.

I am not sure what "literal" means here, but asciidoc doesn't render properly when a back-ticked phrase isn't followed by a space.

garyrussell: Sure; fine with me; we don't talk about how it's mapped for other binders.

PR #506, opened 2018-11-27
Enable KafkaStreams bootstrap tests

### LINKED ISSUE #501: KafkaStreamsBinderBootstrapTest takes longer time to complete
Because it takes so long to complete, currently this test is ignored. Need to investigate why this test takes so long.

sobychacko:
Remove the usage of ImportBeanDefinitionRegistrar in Kafka Streams binder
components since the regular use of getBean from the outer context is safe to do so.

Unignore tests

Resolves #501

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
@@ -19,13 +19,15 @@ package org.springframework.cloud.stream.binder.kafka.streams;
 import java.util.Map;
 
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.config.BeanFactoryPostProcessor;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.kafka.KafkaProperties;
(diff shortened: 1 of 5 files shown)


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KTableBinderConfiguration.java
```diff
@@ -37,13 +36,14 @@
  */
 @SuppressWarnings("ALL")
 @Configuration
-@Import(KafkaStreamsBinderUtils.KafkaStreamsMissingBeansRegistrar.class)
 public class KTableBinderConfiguration {
 
 	@Bean
 	@ConditionalOnBean(name = "outerContext")
 	public BeanFactoryPostProcessor outerContextBeanFactoryPostProcessor() {
```
artembilan: Must be `static`.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinderConfiguration.java
```diff
@@ -34,9 +36,22 @@
  * @since 2.1.0
  */
 @Configuration
-@Import(KafkaStreamsBinderUtils.KafkaStreamsMissingBeansRegistrar.class)
 public class GlobalKTableBinderConfiguration {
 
+	@Bean
+	@ConditionalOnBean(name = "outerContext")
+	public BeanFactoryPostProcessor outerContextBeanFactoryPostProcessor() {
```
artembilan: Must be `static`.

Issue #525, opened 2019-01-03
Support channel specific group name for PollableMessageSource


Walliee:
Currently, `spring.cloud.stream.bindings.<channelName>.group` property doesn't take effect when `PollableMessageSource` are created causing it to fallback to `spring.cloud.stream.kafka.consumer-properties.group.id`

'groupId' needs to be set when `KafkaMessageChannelBinder` creates `KafkaMessageSource`.

Kafka binder version: 2.1.0.RC4

Walliee:
On closer inspection, looks like the issue is with the way `KafkaMessageChannelBinder` creates ConsumerFactory. The channel specific groupId gets overridden by groupId configured at the binder level.

```java 
protected ConsumerFactory<?, ?> createKafkaConsumerFactory(boolean anonymous, String consumerGroup,
			ExtendedConsumerProperties<KafkaConsumerProperties> consumerProperties) {
		Map<String, Object> props = new HashMap<>();
		props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, ByteArrayDeserializer.class);
		props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, ByteArrayDeserializer.class);
		props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, false);
		props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, 100);
		props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, anonymous ? "latest" : "earliest");
		props.put(ConsumerConfig.GROUP_ID_CONFIG, consumerGroup);

		Map<String, Object> mergedConfig = this.configurationProperties.mergedConsumerConfiguration();
		if (!ObjectUtils.isEmpty(mergedConfig)) {
// mergedConfig has ConsumerConfig.GROUP_ID_CONFIG key 
// set to the value of spring.cloud.stream.kafka.consumer-properties.group.id
			props.putAll(mergedConfig);
		}
                ...
}
```

PR #526, opened 2019-01-03
Allow channel specific group names

Walliee:
- use chanel specific group name over binder level group id/name
config
- use binder level group id/name config if channel group name is
anonymous
- use anonymous group name if binder level group id/name config is
absent

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -911,11 +911,13 @@ public class KafkaMessageChannelBinder extends
 		props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, false);
 		props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, 100);
 		props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, anonymous ? "latest" : "earliest");
-		props.put(ConsumerConfig.GROUP_ID_CONFIG, consumerGroup);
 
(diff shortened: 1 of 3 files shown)


Walliee:
Fixes #525

garyrussell:
This is a legitimate bug, the `binder.consumer-properties` group.id should **never** win; it violates the semantics of 

>**group**

>The consumer group of the channel. Applies only to inbound bindings. See Consumer Groups.

>Default: null (indicating an anonymous consumer).

In that if there is no group, the consumer **must** be anonymous. Any boot or binding-level setting for `group.id` must be ignored.

This is certainly wrong...

```java
		props.put(ConsumerConfig.GROUP_ID_CONFIG, consumerGroup);

		Map<String, Object> mergedConfig = this.configurationProperties.mergedConsumerConfiguration();
		if (!ObjectUtils.isEmpty(mergedConfig)) {
			props.putAll(mergedConfig);
		}
```
But we can't move it up, because we want most of the other configuration to be applied (overwrite deserializers for example).

The `KafkaBinderConfigurationProperties.mergedConsumerConfiguration()` should filter out any GROUP_ID_CONFIG property, as well as ENABLE_AUTO_COMMIT_CONFIG because that, too would violate binding properties; we never want the kafka-clients to manage the commits). When filtering out such properties, we might want to log a WARN explaining that the setting should be on the binding instead.

garyrussell:
Furthermore, allowing the GROUP_ID_CONFIG to win will violate other component names (error channel etc, which includes the group name), which would be wrong for anon. consumers.

Walliee:
@garyrussell Thanks for the pointers! What you said makes sense, in fact I thought about it too when I originally created this PR. 

I was concerned however that despite this (group.id sourcing from Boot/SCS consumer properties) being unintentional, there might be consumers relying on existing behaviour.

garyrussell:
>I was concerned 

It's a reasonable concern; we generally try to avoid breaking changes, however an existing app relying on a bug would not generally be a concern.

Let's continue the discussion on #525

Issue #529, opened 2019-01-13
add support for zstd compression


cforce:
https://cwiki.apache.org/confluence/display/KAFKA/KIP-110%3A+Add+Codec+for+ZStandard+Compression
Currently not supported in https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaProducerProperties.java

olegz:
Not sure but perhaps it is more relevant to spring-kafka? @garyrussell ?

garyrussell:
Also, this logic in the provisioner needs to be improved...

```java
try {
	//This call may return null or throw an exception.
	partitions = callable.call();
}
catch (Exception ex) {
	//The above call can potentially throw exceptions such as timeout. If we can determine
	//that the exception was due to an unknown topic on the broker, just simply rethrow that.
	if (ex instanceof UnknownTopicOrPartitionException) {
		throw ex;
	}
}
```

The actual exception is eaten; it should be at least logged.

In this case, the actual `KafkaException` has a cause:

    java.lang.IllegalArgumentException: Unknown compression name: zstd

But at runtime, all we see is...

```
org.springframework.cloud.stream.binder.BinderException: Cannot initialize binder:
	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.getPartitionsForTopic(KafkaTopicProvisioner.java:451) ~[spring-cloud-stream-binder-kafka-core-2.1.0.RC4.jar:2.1.0.RC4]
	at org.springframework.cloud.stream.binder.kafka.KafkaMessageChannelBinder.createProducerMessageHandler(KafkaMessageChannelBinder.java:284) ~[spring-cloud-stream-binder-kafka-2.1.0.RC4.jar:2.1.0.RC4]
	at org.springframework.cloud.stream.binder.kafka.KafkaMessageChannelBinder.createProducerMessageHandler(KafkaMessageChannelBinder.java:132) ~[spring-cloud-stream-binder-kafka-2.1.0.RC4.jar:2.1.0.RC4]
	at org.springframework.cloud.stream.binder.AbstractMessageChannelBinder.doBindProducer(AbstractMessageChannelBinder.java:178) ~[spring-cloud-stream-2.1.0.RC4.jar:2.1.0.RC4]
[...]

PR #530, opened 2019-01-16
GH-529: Add lz4 and docs for zstd compression

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/529

- Also log exception when getting partition information


sobychacko:
Looks good. Merged upstream.

PR #561, opened 2019-03-15
Remove unused setter

### LINKED ISSUE #558: Spring Cloud Stream + Kafka application fails to start when spring.cloud.stream.kafka.binder.required-acks is set to 'all'
The `spring.cloud.stream.kafka.binder.required-acks` property in the `KafkaBinderConfigurationProperties` is impacted by an [issue](https://github.com/spring-projects/spring-boot/issues/16206) in Spring Boot's Binder.

The `acks` property in Kafka producer properties allows for value to be 'all' and to support that `KafkaBinderConfigurationProperties` has the requiredAcks field is of type `String` but along with the regular `String` setter, there is a setter with `int` parameter as well.

https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/02e1fec3b4520125e2f4c6d8d3e79a9c87d1744b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java#L376-L382

This confuses the Boot Binder and the application fails to start up.
Interestingly, it doesn't happen every single time.

I think the setter with the `int` param is not being used anywhere and can be safely removed?

sobychacko:
Removing an integer version of required acks from KafkaBinderConfigurationProperties

Resolves #558

### DIFF
--- a/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
+++ b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
@@ -373,10 +373,6 @@ public class KafkaBinderConfigurationProperties {
 		return this.requiredAcks;
 	}
 
-	public void setRequiredAcks(int requiredAcks) {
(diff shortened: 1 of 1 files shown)


garyrussell:
And cherry-picked to 2.1.x as 2c5dffaa74d3f9430595e1e3e2c1064176abe38f

PR #562, opened 2019-03-16
GH-525: Ignore enable.auto.commit and group.id from merged consumer configuration

### LINKED ISSUE #525: Support channel specific group name for PollableMessageSource

Walliee:
Closed #526 in favour of this PR
Resolves #525

### DIFF
--- a/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
+++ b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
@@ -20,11 +20,14 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import javax.validation.constraints.AssertTrue;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.producer.ProducerConfig;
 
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
```diff
@@ -559,6 +565,17 @@ public void setProducerProperties(Map<String, String> producerProperties) {
 				ProducerConfig.BOOTSTRAP_SERVERS_CONFIG);
 	}
 
+	private void filterStreamManagedConfiguration(Map<String, Object> configuration) {
+		Stream.of(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, ConsumerConfig.GROUP_ID_CONFIG)
+				.forEach(streamManagedConfiguration -> {
+					if (configuration.containsKey(streamManagedConfiguration)) {
+						logger.warn("Ignoring provided value(s) for '" + streamManagedConfiguration + "'. " +
+								"This configuration (and related functionality) is managed by Spring Cloud Stream.");
+						configuration.remove(streamManagedConfiguration);
```
garyrussell: Given that some might consider this a breaking change (although we generally don't consider a fix that breaks code that relyies on incorrect behavior [bug] as "breaking"), I think we need more specific warning logs.

I also think we should only warn if they explicitly set auto-commit to 'true', setting it to false is ok.

`ConsumerConfig. ENABLE_AUTO_COMMIT_CONFIG + "=true is not supported by the Kafka binder`
`Use spring.cloud.stream.binding.<name>.group to specify the group instead of " + ConsumerConfig.GROUP_ID_CONFIG`

sabbyanandan:
@Walliee: Any thoughts on @garyrussell comment?

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
Walliee: @garyrussell I was trying to keep the warning generic just in case more configuration is added to the set of ignored/overridden configuration but I've now updated the PR with more explicit warnings based on your suggestions.

PR #637, opened 2019-04-12
Kafka Streams multiple function issues

### LINKED ISSUE #636: Kafka Streams multiple function issues
When there are multiple functions provided as `java.util.function.Function` or `java.util.function.Consumer` beans, there are few issues in the current function support in the Kafka Streams binder.

sobychacko:
Fixing a bug around when we have muliple beans defined as functions/csonumers in the new
Kafka Streams binder functional support.

Polishing

Resolves #636

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/FunctionDetectorCondition.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/FunctionDetectorCondition.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2019-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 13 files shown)


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/FunctionDetectorCondition.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class FunctionDetectorCondition extends SpringBootCondition {
+
+	@Override
+	public ConditionOutcome getMatchOutcome(ConditionContext context, AnnotatedTypeMetadata metadata) {
+		if (context != null &&  context.getBeanFactory() != null) {
+			final Map<String, Function> functionTypes = context.getBeanFactory().getBeansOfType(Function.class);
+			final Map<String, Consumer> consumerTypes = context.getBeanFactory().getBeansOfType(Consumer.class);
+
+			if (!functionTypes.isEmpty() || !consumerTypes.isEmpty()) {
```
olegz: Since you no longer rely on `definition` property and instead check for existence, what would happen in the event I have a Function bean that was not meant to be wired for streams? Perhaps you can use FunctionInspector and actually check for signature and see if those functions are KStream. This way you (at least) narrow the scope.

PR #643, opened 2019-04-22
Custom state stores with functional Kafka Streams

sobychacko:
Only merge after https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/637

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsFunctionProcessor.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsFunctionProcessor.java
@@ -71,6 +71,7 @@ import org.springframework.util.StringUtils;
 
 /**
  * @author Soby Chacko
+ * @since 2.2.0
  */
 public class KafkaStreamsFunctionProcessor implements ApplicationContextAware {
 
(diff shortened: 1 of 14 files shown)


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsFunctionProcessor.java
```diff
@@ -151,11 +154,22 @@ public void orchestrateFunctionInvoking(ResolvableType resolvableType, String fu
 		Object[] adaptedInboundArguments = adaptAndRetrieveInboundArguments(stringResolvableTypeMap, functionName);
 		try {
 			if (resolvableType.getRawClass() != null && resolvableType.getRawClass().equals(Consumer.class)) {
+				//TOOD: Investigate why looking up by Consumer returns null
```
olegz: Soby, there is nothing to investigate. It's a SCF issue that we simply need to document. Basically you can never have Consumer since after wrapping (which is what's in FunctionCatalog) Functions and Consumers are all Functions. So consumer is `Function<Flux<?>, Mono<Void>>`

olegz: So you can technically remove this whole block of code

PR #669, opened 2019-06-11
Use native Serde for Kafka Streams binder as the default

### LINKED ISSUE #651: Consider making native serialization/deserialization the default in Kafka Streams binder
Kafka streams applications most often use its own `Serde` mechanism to do input deserialization and output serialization. However, Kafka Streams binder in `ScSt` uses message converters provided by the framework as the default serialization strategy. There are use cases in which an application must use the framework provided serialization (for e.g. when the incoming data is serialized with a particular serializer that is not compatible with any Serde's provided by Kafka). However, this is not very common. Therefore, I propose, in the 2.3 version of the binder, we make native de/serialization as the default serialization strategy with the option given for the users to change to framework serialization if need be. 

Currently we set both `useNativeDeocding` and `useNativeEncoding` to `false` in the binder. With this change, they default to `true`.  The downside is that the applications must provide a `Serde` for both key/value in the application's configuration.

sobychacko:
In Kafka Streams binder, use the native Serde mechanism as the default instead of the framework
provided message conversion on the input and output bindings. This is to align applications
written using Kafka Streams binder more compatible with native concepts and mechanisms.
Users can still disable native encoding and decoding through configuration.

Amend tests to accommodate the flipped configuration.

Resolves #651

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KStreamBoundElementFactory.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KStreamBoundElementFactory.java
@@ -22,6 +22,7 @@ import org.apache.kafka.streams.kstream.KStream;
 
 import org.springframework.aop.framework.ProxyFactory;
 import org.springframework.cloud.stream.binder.ConsumerProperties;
+import org.springframework.cloud.stream.binder.ProducerProperties;
 import org.springframework.cloud.stream.binding.AbstractBindingTargetFactory;
 import org.springframework.cloud.stream.config.BindingProperties;
 import org.springframework.cloud.stream.config.BindingServiceProperties;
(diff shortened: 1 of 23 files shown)


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KStreamBoundElementFactory.java
```diff
@@ -52,8 +53,12 @@
...
 	@Override
 	public KStream createInput(String name) {
-		ConsumerProperties consumerProperties = this.bindingServiceProperties
-				.getConsumerProperties(name);
+		BindingProperties bindingProperties = this.bindingServiceProperties.getBindingProperties(name);
+		ConsumerProperties consumerProperties = bindingProperties.getConsumer();
+		if (consumerProperties == null) {
+			consumerProperties = this.bindingServiceProperties.getConsumerProperties(name);
+			consumerProperties.setUseNativeDecoding(true);
+		}
```
olegz: This code here appears to be identical to the one in `GlobalKTableBoundElementFactory.createInput(..)` and `KTableBoundElementFactory.createInput(..)`. Have you considered some abstract class?

sobychacko: sure, I am going to refactor a bunch of these classes. Can I address these concerns as part of that?

PR #672, opened 2019-06-13
Infer SerDes used on input/output when possible

### LINKED ISSUE #368: Infer SerDes used on input/output when possible
It will be helpful if the framework can infer the Serde's used on input and output whenever possible (aka: when native SerDes are enabled). This way, the users don't have to specify the Serdes as properties if they are different from the default Serdes.

sobychacko:
When using native de/serializaion (which is now the default), the binder should infer the common Serde's
used on input and output. The Serdes inferred are - Integer, Long, Short, Double, Float, String, byte[] and
Spring Kafka provided JsonSerde.

Resolves #368

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
@@ -47,7 +47,7 @@ class KafkaStreamsBindingInformationCatalogue {
 
 	private final Set<StreamsBuilderFactoryBean> streamsBuilderFactoryBeans = new HashSet<>();
 
-	private ResolvableType outboundResolvable;
+	private ResolvableType outboundKStreamResolvable;
 
 	/**
 	 * For a given bounded {@link KStream}, retrieve it's corresponding destination on the
@@ -125,11 +125,11 @@ class KafkaStreamsBindingInformationCatalogue {
 		return this.streamsBuilderFactoryBeans;
 	}
 
-	public void addOutputKStreamResolvable(ResolvableType outboundResolvable) {
-		this.outboundResolvable = outboundResolvable;
+	public void setOutboundKStreamResolvable(ResolvableType outboundResolvable) {
+		this.outboundKStreamResolvable = outboundResolvable;
 	}
 
-	public ResolvableType getOutboundResolvable() {
-		return outboundResolvable;
+	public ResolvableType getOutboundKStreamResolvable() {
+		return outboundKStreamResolvable;
 	}
 }
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
@@ -27,6 +27,7 @@ import org.apache.kafka.streams.kstream.KStream;
 import org.springframework.cloud.stream.binder.ConsumerProperties;
 import org.springframework.cloud.stream.binder.kafka.streams.properties.KafkaStreamsConsumerProperties;
 import org.springframework.cloud.stream.config.BindingProperties;
+import org.springframework.core.ResolvableType;
 import org.springframework.kafka.config.StreamsBuilderFactoryBean;
 
 /**
@@ -45,6 +46,9 @@ class KafkaStreamsBindingInformationCatalogue {
 	private final Map<KStream<?, ?>, KafkaStreamsConsumerProperties> consumerProperties = new ConcurrentHashMap<>();
 
 	private final Set<StreamsBuilderFactoryBean> streamsBuilderFactoryBeans = new HashSet<>();
+
+	private Map<KStream, ResolvableType> outputBindings;
+	private ResolvableType outboundResolvable;
 
 	/**
 	 * For a given bounded {@link KStream}, retrieve it's corresponding destination on the
@@ -122,4 +126,19 @@ class KafkaStreamsBindingInformationCatalogue {
 		return this.streamsBuilderFactoryBeans;
 	}
 
+	public void addThisMap(Map<KStream, ResolvableType> outputBindings) {
+		this.outputBindings = outputBindings;
+	}
+
+	public Map<KStream, ResolvableType> getOutputBindings() {
+		return outputBindings;
+	}
+
+	public void addOutputKStreamResolvable(ResolvableType outboundResolvable) {
+		this.outboundResolvable = outboundResolvable;
+	}
+
+	public ResolvableType getOutboundResolvable() {
+		return outboundResolvable;
+	}
 }
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
@@ -47,7 +47,6 @@ class KafkaStreamsBindingInformationCatalogue {
 
 	private final Set<StreamsBuilderFactoryBean> streamsBuilderFactoryBeans = new HashSet<>();
 
-	private Map<KStream, ResolvableType> outputBindings;
 	private ResolvableType outboundResolvable;
 
 	/**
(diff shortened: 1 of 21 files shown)


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
```diff
@@ -122,4 +126,19 @@ void addStreamBuilderFactory(StreamsBuilderFactoryBean streamsBuilderFactoryBean
 		return this.streamsBuilderFactoryBeans;
 	}
 
+	public void addThisMap(Map<KStream, ResolvableType> outputBindings) {
```
olegz: What is `thisMap`?

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
```diff
@@ -122,4 +126,19 @@ void addStreamBuilderFactory(StreamsBuilderFactoryBean streamsBuilderFactoryBean
...
+
+	public Map<KStream, ResolvableType> getOutputBindings() {
+		return outputBindings;
+	}
+
+	public void addOutputKStreamResolvable(ResolvableType outboundResolvable) {
+		this.outboundResolvable = outboundResolvable;
+	}
+
+	public ResolvableType getOutboundResolvable() {
```
olegz: Few lines above you have a method `add. . . ` which is different than `set. . .` since `add` implies some type of collection, and here you are executing `get. . .`. Perhaps a bit more cleanup

olegz: Also, would be nice if get/set have some matching names

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsFunctionProcessor.java
```diff
@@ -184,6 +189,8 @@ public void orchestrateFunctionInvoking(ResolvableType resolvableType, String fu
 					i++;
 				}
 				if (result != null) {
+					kafkaStreamsBindingInformationCatalogue.addOutputKStreamResolvable(
+							outboundResolvableType != null ? outboundResolvableType : resolvableType.getGeneric(1));
```
olegz: I am assuming `resolvableType.getGeneric(1)` can't be null or setting  `addOutputKStreamResolvable` to null is ok

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsFunctionProcessor.java
```diff
@@ -251,9 +258,19 @@ public void orchestrateFunctionInvoking(ResolvableType resolvableType, String fu
...
+					Serde<?> valueSerde;
+
+					if (bindingServiceProperties.getConsumerProperties(input).isUseNativeDecoding()) {
+						keySerde = this.keyValueSerdeResolver.getInboundKeySerde(extendedConsumerProperties, stringResolvableTypeMap.get(input));
+						valueSerde = this.keyValueSerdeResolver.getInboundValueSerde(
+								bindingProperties.getConsumer(), extendedConsumerProperties, stringResolvableTypeMap.get(input));
+					}
+					else {
+						keySerde = Serdes.ByteArray();
+						valueSerde = Serdes.ByteArray();
```
olegz: Does it have to be separate instance of Serdes.ByteArray() for key and value? Could it be shared between the two?
```
Serde serde = Serdes.ByteArray();
Serde<?> keySerde = serde;
Serde<?> valueSerde = serde;
if (bindingServiceProperties.getConsumerProperties(input). . . ) {
. . .
}
```

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsStreamListenerSetupMethodOrchestrator.java
```diff
@@ -268,10 +271,21 @@ else if (arguments.length == 1 && StringUtils.hasText(inboundName)) {
...
+					if (bindingServiceProperties.getConsumerProperties(inboundName).isUseNativeDecoding()) {
+
+						keySerde = this.keyValueSerdeResolver
+								.getInboundKeySerde(extendedConsumerProperties, ResolvableType.forMethodParameter(methodParameter));
+						valueSerde = this.keyValueSerdeResolver.getInboundValueSerde(
+								bindingProperties.getConsumer(), extendedConsumerProperties, ResolvableType.forMethodParameter(methodParameter));
+					}
+					else {
+						keySerde = Serdes.ByteArray();
+						valueSerde = Serdes.ByteArray();
```
olegz: Same serde comment as well as the fact that the code looks similar to the code above which may be better to be moved into something shared (e.g., utility class, abstract class etc)

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KeyValueSerdeResolver.java
```diff
@@ -140,6 +178,25 @@
...
+			if (producerProperties.isUseNativeEncoding()) {
+				valueSerde = getValueSerde(
+						kafkaStreamsProducerProperties.getValueSerde(), resolvableType);
+			}
+			else {
+				valueSerde = Serdes.ByteArray();
+			}
+			valueSerde.configure(this.streamConfigGlobalProperties, false);
+		}
+		catch (ClassNotFoundException ex) {
```
olegz: Given the possibility of CNFE I am assuming there is some class loading voodoo here, but I don't see which class loader is used. Just think if in this context it could be a problem

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
sobychacko: Removed this.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
sobychacko: done.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsFunctionProcessor.java
sobychacko: ya, that's right, in most cases, the `outboundResolvableType` is already set by the pre-processors. Otherwise, we get that from the functions' return signature.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsFunctionProcessor.java
sobychacko: Changed the logic a little bit here in the new commits.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsStreamListenerSetupMethodOrchestrator.java
sobychacko: yep, this is going to be cleaned up as part of a refactoring PR after this.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KeyValueSerdeResolver.java
sobychacko: ya, this CNFE is thrown by one of the Kafka classes and we are catching that and re-throw as an `IllegalStateException` with the original exception wrapped.

Issue #423, opened 2018-08-01
KafkaMessageChannelBinder should support setting the topic name per message through the use of KafkaHeaders.TOPIC (i.e. "kafka_topic" header)


ngbalk:
The `KafkaProducerMessageHandler` has the following logic to allow for the producer to drive the topic name through setting the "kafka_topic" header:

```java
String topic = this.topicExpression != null ?
		this.topicExpression.getValue(this.evaluationContext, message, String.class)
		: messageHeaders.get(KafkaHeaders.TOPIC, String.class);
```

However, in the `KafkaMessageChannelBinder`, we always set the topic expression:
https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/7007f9494a2f43e066141dda90c85cde62a83f1c/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L878

I feel that it would be nice if we could somehow take advantage of this feature in spring-integration-kafka by providing some logic or configuration to optionally set the topicExpression.  Then we could use the "kafka_topic" field to drive the destination topic name.  

I've noticed that the spring-cloud-streams abstraction will try to use the channelName as the topic name, so maybe we need some property like `headerDrivenTopicResolution=true` that will skip setting the topicExpression and rely on the "kafka_topic" header for the topic name (and fail if not present).

Thoughts?

Nick

sabbyanandan:
Hi, @ngbalk / @Walliee! Based on our discussion yesterday, it sounded like you're already doing this using `BinderAwareChannelResolver`. You may have already managed to solve it with customizations - did I get that right? 

It'd be nice to learn how you're doing it, and if it is generic enough, we could even add it to the framework. 

If I misinterpreted any of this, feel free to clarify with a use-case, and we will be open to exploring options on our side.

ngbalk:
@sabbyanandan  We use `BinderAwareChannelResolver` along with `AbstractMappingMessageRouter` to achieve our "multi-tenant topic routing".  In pseudo-code, it ends up looking something like this in our Impl of `AbstractMappingMessageRouter` : 
```java

  @Override
  protected List<Object> getChannelKeys(Message<?> message) {
    String tenantId = determineTenantId();
    String destinationTopicName = tenantId + "_" + originalTopicName;
    return destinationTopicName;
  }
```
[...]

PR #679, opened 2019-06-18
GH-423 Add option to use KafkaHeaders.TOPIC Header

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/423

Add an option to override the default binding destination with the value
of the header, if present.


sobychacko:
Merged upstream.

PR #692, opened 2019-06-30
Functional enhancements in Kafka Streams binder [DO NOT MERGE-ONLY FOR REVIEW]

sobychacko:


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/KafkaStreamsBindableProxyFactory.java
```diff
@@ -0,0 +1,257 @@
...
+ * @since 3.0.0
+ */
+public class KafkaStreamsBindableProxyFactory extends AbstractBindableProxyFactory implements InitializingBean, BeanFactoryAware {
+
+	/**
+	 * Default output binding name. Output binding may occur later on in the function invoker (outside of this class),
+	 * thus making this field part of the API.
+	 */
+	public static final String DEFAULT_OUTPUT_SUFFIX = "out";
+	private static final String DEFAULT_INPUT_SUFFIX = "in";
```
olegz: At some point before releasing 3.0 we will move these constants to the core

PR #700, opened 2019-07-03
GH-694: Add recordMetadataChannel (send confirm.)

### LINKED ISSUE #694: Add support to provide a mechanism to track Send Result Metadata
@NLatyshev commented on [Thu Jun 27 2019](https://github.com/spring-cloud/spring-cloud-stream/issues/1750)

Hi everyone.

Is there way to know if message sent or not?

For example, I have Kafka binder and use MessageChannel to produce message to some topic (e.g. in some controller). Method send always returns true (and I didn't found any other method to pass callback) even if Kafka connection lost. So seems user newer find out he/she request is lost.

Can someone suggest resolution? Is it documentation defect or just the functionality is absent (and not planned)

garyrussell:
Resolves #694

Also fix deprecation of `ChannelInterceptorAware`.

### DIFF
--- a/docs/src/main/asciidoc/overview.adoc
+++ b/docs/src/main/asciidoc/overview.adoc
@@ -332,6 +332,16 @@ Set to `true` to override the default binding destination (topic name) with the 
 If the header is not present, the default binding destination is used.
 Default: `false`.
 +
+recordMetadataChannel::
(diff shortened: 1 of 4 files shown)


sobychacko:
Merged upstream.

PR #753, opened 2019-09-25
Check HeaderMapper bean with a well known name

### LINKED ISSUE #749: Consider making the default `DefaultKafkaHeaderMapper` a bean
in Sleuth we'll need to add wrapping of `DefaultKafkaHeaderMapper` implementations. Unfortunately the default `DefaultKafkaHeaderMapper` object is not a bean so Sleuth can't wrap it.

It would be great to consider making the default `DefaultKafkaHeaderMapper` instance as a bean.

related issue https://github.com/spring-cloud/spring-cloud-sleuth/issues/1430

sobychacko:
* If a custom bean name for header mapper is not provided through the binder property headerMapperBeanName,
  then look for a header mapper bean with the name kafkaHeaderMapper.
* Add tests to verify

Resolves #749

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -388,7 +388,7 @@ public class KafkaMessageChannelBinder extends
 		if (mapper == null) {
 			//First, try to see if there is a bean named headerMapper registered by other frameworks using the binder (for e.g. spring cloud sleuth)
 			try {
-				mapper = getApplicationContext().getBean("kafkaHeaderMapper", KafkaHeaderMapper.class);
+				mapper = getApplicationContext().getBean("kafkaBinderHeaderMapper", KafkaHeaderMapper.class);
 			}
 			catch (BeansException be) {
 				// Pass through
@@ -913,7 +913,7 @@ public class KafkaMessageChannelBinder extends
 		if (mapper == null) {
 			//First, try to see if there is a bean named headerMapper registered by other frameworks using the binder (for e.g. spring cloud sleuth)
 			try {
-				mapper = getApplicationContext().getBean("kafkaHeaderMapper", KafkaHeaderMapper.class);
+				mapper = getApplicationContext().getBean("kafkaBinderHeaderMapper", KafkaHeaderMapper.class);
 			}
 			catch (BeansException be) {
 				DefaultKafkaHeaderMapper headerMapper = new DefaultKafkaHeaderMapper() {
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -325,8 +328,6 @@ public void testTrustedPackages() throws Exception {
 				producerBindingProperties);
 
 		ExtendedConsumerProperties<KafkaConsumerProperties> consumerProperties = createConsumerProperties();
-		consumerProperties.getExtension()
-				.setTrustedPackages(new String[] { "org.springframework.util" });
```
garyrussell: Why do we no longer need to test this property?

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -377,12 +378,23 @@ protected MessageHandler createProducerMessageHandler(
...
 		KafkaHeaderMapper mapper = null;
 		if (this.configurationProperties.getHeaderMapperBeanName() != null) {
 			mapper = getApplicationContext().getBean(
 					this.configurationProperties.getHeaderMapperBeanName(),
 					KafkaHeaderMapper.class);
 		}
+		if (mapper == null) {
+			//First, try to see if there is a bean named headerMapper registered by other frameworks using the binder (for e.g. spring cloud sleuth)
+			try {
+				mapper = getApplicationContext().getBean("kafkaHeaderMapper", KafkaHeaderMapper.class);
```
garyrussell: I think this name should have something to do with the project. `kafkaBinderHeaderMapper` ?

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
sobychacko: Actually, we were not doing any assertions on that. I will add it back and assert.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
sobychacko: Good idea. Will update.

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
sobychacko: Sorry, my mistake. We were doing assertions on the mimetype from that package. Will restore.

PR #762, opened 2019-10-03
Kafka Streams DLQ enhancements

### LINKED ISSUE #761: Kafka Streams binder DLQ enhancements
Since Spring Kafka now supports [DLQ](https://docs.spring.io/spring-kafka/docs/2.3.0.RELEASE/reference/html/#streams-deser-recovery) through `DeadLetterPublishingRecoverer`, the binder should use that instead of custom DLQ components.

sobychacko:
Use DeadLetterPublishingRecoverer from Spring Kafka instead of custom DLQ components in the binder.
Remove code that is no longer needed for DLQ purposes.
In Kafka Streams, always set group to application id if the user doesn't set an explicit group.

Upgrade Spring Kafka to 2.3.0 and SIK to 3.2.0

Resolves #761


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderUtils.java
```diff
@@ -59,33 +70,73 @@ static void prepareConsumerBinding(String name, String group,
...
+			DeadLetterPublishingRecoverer kafkaStreamsBinderDlqRecoverer = !StringUtils
 					.isEmpty(extendedConsumerProperties.getExtension().getDlqName())
-							? new KafkaStreamsDlqDispatch(
-									extendedConsumerProperties.getExtension()
-											.getDlqName(),
-									binderConfigurationProperties,
-									extendedConsumerProperties.getExtension())
-							: null;
+					? new DeadLetterPublishingRecoverer(kafkaTemplate, (cr, e) -> new TopicPartition(extendedConsumerProperties.getExtension()
+					.getDlqName(), cr.partition()))
```
garyrussell: With this logic, the DLT must have the same number of partitions as the original.

We should probably add a new property/properties to allow the user to determine whether the DLQ publish should be done to the same partition, a particular partition, or no partition (kafka chooses).

sobychacko: @garyrussell We will address this as a separate issue (maybe after S1P) - Filed [this issue](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/763).

Issue #763, opened 2019-10-03
Kafka Streams DLQ - more fine grained partitions control


sobychacko:
Right now in Kafka Streams binder, it expects that the DLQ has the same number of partitions as the input topic. That is true if the provisioner creates the topic, but if it doesn't, then the app needs to ensure that the DLQ is created with the right number of partitions. In order to address this, we should make the creation of DLQ topic more flexible. Maybe something like dlqPartitionExpression evaluated against the CR, with the default being cr.partition(); if it evaluates to null, Kafka will choose the partition, or the expression could simply evaluate to 0, 1, 2, etc.

y-manish:
Yeah, having control on DLQ partition count will be helpful. Thank you.
For now, is there some work around way for this as of now(version : 2.1.x).

garyrussell:
There is no work around; sorry.

PR #778, opened 2019-10-22
GH-763: Add DlqPartitionFunction

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/763

Allow users to select the DLQ partition based on

* consumer group
* consumer record (which includes the original topic/partition)
* exception


sobychacko:
LGTM - Merged upstream.

Issue #628, opened 2019-04-08
Allow for custom partition count in DLQ


benjvoigt:
According to [this SO answer from @garyrussell](https://stackoverflow.com/a/55364072/987860) it is currently not possible to configure a divergent partition count than the origin topic for DLQs. Since we do not require as many partitions for the DLQs and kafka allocates a good amount of disk space for each of them we'd like to have it set to some lower value.

spring-kafka does support this feature already by giving the option to provide a custom `DeadLetterPublishingRecoverer` that implements `BiFunction<ConsumerRecord<?, ?>, Exception, TopicPartition>`, read https://docs.spring.io/spring-kafka/reference/html/#dead-letters.

For us it would be sufficient to get a configurable property along the lines `spring.cloud.stream.kafka.bindings.<channelName>.consumer.dlqPartitionCount` which implies a new partition being set for each message that is send to the dlq.

fr-manish:
@garyrussell , Thank you.
By any chance, will it be back ported to other older version(v2.1.x) as well ? 
It would be helpful to get this there.

PR #780, opened 2019-10-23
GH-628: Add dlqPartitions property

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/628

Allow users to specify the number of partitions in the dead letter topic.

If the property is set to 1 and the `minPartitionCount` binder property is 1,
override the default behavior and always publish to partition 0.


sobychacko:
Looks great. Merged upstream.

Issue #715, opened 2019-08-08
Binder DLQ/Retry cannot be used with transactions


garyrussell:
See [Stack Overflow](https://stackoverflow.com/questions/57379575/spring-cloud-stream-multi-topic-transaction-management/57379903#57379903).

When using transactions, you can't perform retry and DLQ publishing at the binder level.

Consider `consume->produce->throw exception`.

Binder level retry/DLQ absorbs the exception and the transaction is not rolled back, resulting in unexpected published messages (multiple with retry).

This can be avoided by disabling retry/dlq in the binder and configuring the container with an appropriately configured `AfterRollbackProcessor` (see the  stack overflow answer above).

However retry backoff is not currently supported  by the `DefaultAfterRollbackProcessor` - the retries will be immediate; there is a [spring-kafka PR to add backoff](https://github.com/spring-projects/spring-kafka/pull/1185).

It should be possible to use the binder configuration to configure the after rollback processor automatically when the binder is transactional, once that PR is merged.

PR #799, opened 2019-11-12
GH-715: Add retry, dlq for transactional binders

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/715

When using transactions, binder retry, dlq cannot be used because the retry
runs within the transaction, which is undesirable if there is another resource
involved; also publishing to the DLQ could be rolled back.

Use the retry properties to configure an `AfterRollbackProcessor` to perform the
retry and DLQ publishing after the transaction has rolled back.


sobychacko:
Merged upstream.

Issue #739, opened 2019-09-06
KafkaBindingProperties has no documentation (so no popup help in IDE)


dsyer:
Also related nested objects: `KafkaConsumerProperties` and `KafkaProducerProperties`.

sobychacko:
@dsyer `KafkaBindingProperties` is a pojo used in a map from `KafkaExtendddBindingProperties` [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaExtendedBindingProperties.java#L46). Per @snicoll map of pojo won't get metadata support in IDE, but there is an issue to support that in boot. It lists the available properties, but no pop-up docs help. Or are you referring to something else here? Thanks!

dsyer:
I see them in the IDE with "no description". That suggests to me that the IDE would happily add a description if there was one. Whether the AP can generate them is another question (but I don't really see why not).

snicoll:
There never was documentation for Maps of POJO, see https://github.com/spring-projects/spring-boot/issues/9945

dsyer:
Maybe the IDE would display Javadocs if they were available in the POJO (that's what the popup in Eclipse suggests to me at least)? The properties need to be documented IMO anyway, so it's worth doing. The only question is whether to do it in a format acceptable to the AP, in case one day it decides to support this as well.

dsyer:
OK, so I tested it. Eclipse (STS) will show the Javadocs from the getter method (not the field) in a POJO. I think we should put some docs in there. And maybe when the AP catches up it can look there as well.

sobychacko:
@dsyer Is it only for the POJO's inside the map that Javadocs need to be added on the getter methods and not on the fields? We usually add Javadocs on the fields only.

dsyer:
Yes. Fields is what the AP supports. But the AP does not support mapped POJOs currently.

PR #802, opened 2019-11-13
KafkaBindingProperties has no documentation

sobychacko:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/739

No popup help in IDE due to missing javadocs in KafkaBindingProperties,
KafkaConsumerProperties and KafkaProducerProperties.

Some IDEs currently only show javadocs on getter methods from POJOs
used insdide a map. Therefore, some javadocs are duplicated between
fields and getter methods.

### DIFF
--- a/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaConsumerProperties.java
+++ b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaConsumerProperties.java
@@ -84,40 +84,108 @@ public class KafkaConsumerProperties {
 
 	}
 
+	/**
+	 * When autoCommitOffset is true, this setting dictates whether to commit the offset after each record is processed.
+	 */
 	private boolean ackEachRecord;
 
+	/**
+	 * When true, topic partitions is automatically rebalanced between the members of a consumer group.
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaConsumerProperties.java
```diff
@@ -84,47 +84,120 @@ public long getReferencePoint() {
 
 	}
 
+	/**
+	 * When autoCommitOffset is true, this setting dictates whether to commit the offset after each record is processed.
```
garyrussell: instead of after the complete batch of records returned by a poll() is processed.

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaConsumerProperties.java
```diff
@@ -85,7 +85,8 @@ public long getReferencePoint() {
 	}
 
 	/**
-	 * When autoCommitOffset is true, this setting dictates whether to commit the offset after each record is processed.
+	 * When autoCommitOffset is true, this setting dictates whether to commit the offset
+	 * after the complete batch of records returned by a poll() is processed.
```
garyrussell: This doesn't reflect the change I suggested.

Perhaps - when `true` the offset is committed after each record, otherwise the offsets for the complete set of records received from the poll() are committed after all records have been processed.

PR #831, opened 2020-01-16
Enable config of authorizationExceptionRetryInterval through properties

### LINKED ISSUE #830: Enable usage of spring-kafka's authorizationExceptionRetryInterval
Spring-kafka 2.3.5 introduces authorizationExceptionRetryInterval parameter (see [spring-kafka/GH-1336](https://github.com/spring-projects/spring-kafka/issues/1336)) to enable retrying in case of authorization errors.

There should be a configuration parameter in Spring Cloud Stream Binder Kafka for this.

lukasz-kaminski:
Resolves #830

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -159,6 +159,13 @@ Use this, for example, if you wish to customize the trusted packages in a `Defau
 +
 Default: none.
 
+spring.cloud.stream.kafka.binder.authorizationExceptionRetryInterval::
+Enables retrying in case of authorization exceptions.
+Defines interval between each retry.
+Accepts `Duration`, e.g. `30s`, `2m`, etc.
++
+Default: `null` (retries disabled, fail fast)
+
 [[kafka-consumer-properties]]
 ==== Kafka Consumer Properties
 
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
 	</parent>
 	<properties>
 		<java.version>1.8</java.version>
-		<spring-kafka.version>2.3.2.RELEASE</spring-kafka.version>
+		<spring-kafka.version>2.3.5.RELEASE</spring-kafka.version>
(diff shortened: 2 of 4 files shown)


sobychacko:
@lukasz-kaminski It might be nice to know the use case that you faced that needed this change in the binder? If it's a common use case that generally applicable, then I think we can consider adding this. Otherwise, as @garyrussell said, using the customizer might be the way to go.

anderseknert:
Our use case is this - we are running Kafka clusters with lots of clients inside each of them. Some topics are meant to be "public" more or less (all clients may consume or produce to them) while others may only be accessed by specific clients under specific conditions. The requirements are not static either - what a client is able to access must be determined if not with each request then at least within an hour or so. On top of this we are often dealing with very sensitive data. For this purpose we have built and installed an [authorization plugin for Kafka](https://github.com/Bisnode/opa-kafka-plugin), which does pretty much exactly this. While initial testing was successful we noticed our Spring based clients ending up in unrecoverable states as soon as one of them got an unauthorized response. We are very happy to see this fixed in the "regular" Spring client as this covers about half of our users - there is still a quite a sizable group using Spring Cloud Stream and from what I understand more are looking at getting onboard.

Whether our use case is common I don't know (it's probably not) - I would however argue that the situation could be worse in more static configurations - after all our setup triggered these problems in our test environments after only a few hours, and that helped us quickly deduce what was likely the problem. Had we run this in a more static setup we would likely have been able to run for weeks or months in production before clients started dying, and we probably wouldn't have a clue why unless we'd spend a considerable effort on troubleshooting.

For our case (me and @lukasz-kaminski) we could certainly get by setting this configuration programmatically - having it exposed as a configuration option would however increase the likeliness of others getting a chance to consider this at an earlier stage.

sobychacko:
@anderseknert Thank you for the detailed explanation. It sounds reasonable.
@lukasz-kaminski The PR is merged upstream. Thank you for contributing.

Issue #836, opened 2020-02-02
Value serde not properly initialized when using @KafkaStreamsStateStore


adrianotr:
I've noticed an issue when using the `@KafkaStreamsStateStore` annotation, if you use `SpecificAvroSerde` as your `valueSerde`.

Apparently the `KeyValueSerdeResolver.getValueSerde` method does not call the `Serde.configure` method to initialize the Serde object after its creation, unlike the `KeyValueSerdeResolver.getKeySerde` method.

As a result, the value Serde does not have the right Schema Registry url.

adrianotr:
I can create a PR for this, unless there's was a reason for the value serde not to be initialized as opposed to the key serde. Please let me know.

olegz:
@adrianotr you should always create a PR since it would be much simpler to reason about it, polish, improve etc. So yes please do it.

PR #840, opened 2020-02-06
Always call value serde configure method

adrianotr:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/836

The value serde was being configured in the wrong place, which was skipping it for state stores. Moved the call to the `getValueSerde` method to make sure it's always configured.


sobychacko:
@adrianotr Did you test with a Kafka Streams app and verified that you see the expected behavior with the `configure` call?

adrianotr:
@sobychacko 
Yes. I built the artifact and replaced it in my project.
I also placed some breakpoints during app initialization and made calls to the methods affected to verify if the schema registry url was set in the schemaRegistryClient inside the SpecificAvroSerde.
Because the `private getValue` method is always called it is the best place to initialize the serde. In fact, that is how the key serde is initialized.

sobychacko:
@adrianotr Thank you for identifying this issue and contribute a fix for this. The PR is now merged upstream after squashing and rebasing the commits. More PR contributions are welcomed!

PR #854, opened 2020-02-26
#833 Add InteractiveQueryService.getAllHostsInfo 

### LINKED ISSUE #833: Make InteractiveQueryService be able to return all metadata for specific store
The current implementation of InteractiveQueryService provides only one method to retrieve metadata for a specific key, but there is no easy wat to retrieve metadata ща all hosts for a specific store.

It will be useful for cases when we should return data for all or some range of keys from state store without moving this logic out of the service.

It would be great to have something like this.

```java
public class InteractiveQueryService { 
 // ...

/**
 * Gets the list of {@link HostInfo} where the provided store is hosted on. 
 * It also can include current host info.  
 * Kafka Streams will look through all the consumer instances under the same application id 
 * and retrieves all hosts info
 *
 * Note that the end-user applications must provide `applicaiton.server` as a configuration property
 * for all the application instances when calling this method. If this is not available, then the empty list will be returned.
 *
 * @param store store name
 * @return the list of {@link HostInfo} where provided store is hosted on
 */
public List<HostInfo> getAllHostsInfo(String store){
    return kafkaStreamsRegistry.getKafkaStreams()
            .stream()
            .flatMap(k -> k.allMetadataForStore(store).stream())
            .filter(Objects::nonNull)
            .map(StreamsMetadata::hostInfo)
            .collect(Collectors.toUnmodifiableList())
}

```

sserhii-95:
Add InteractiveQueryService.getAllHostsInfo to fetch metadata of all instances that handles specific store

Resolves #833

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
@@ -45,6 +45,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Soby Chacko
  * @author Renwei Han
+ * @author Serhii Siryi
  * @since 2.1.0
  */
 public class InteractiveQueryService {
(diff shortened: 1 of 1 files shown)


sserhii-95:
https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/833

sobychacko:
@sserhii-95 Thank you for the PR. I am getting some compile errors (imports missing etc.). Please build it on your end and ensure that the PR branch is clean. Also, any chance you can add a test to verify this new API? There is already [this](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsInteractiveQueryIntegrationTests.java). You need to somehow test the new method here.

sobychacko:
Cleaned up the issues with the PR and added a basic test. Merged the PR upstream into master and back ported to 3.0.x. Thank you for contributing the PR.

Issue #853, opened 2020-02-25
Do not propagate "internal" headers to output dest


garyrussell:
* `deliveryAttempt`
* `scst_nativeHeadersPresent`

PR #855, opened 2020-02-27
GH-853: Don't propagate out "internal" headers

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/853


sobychacko:
Merged upstream.

Issue #870, opened 2020-03-20
Using DLQ in Spring Cloud Stream Binder Kafka does not commit offsets


jskim1991:
Based on the Spring Cloud Stream Binder Kafka documentation:
>autoCommitOnError
Effective only if autoCommitOffset is set to true. If set to false it suppresses auto-commits for messages that result in errors, and will commit only for successful messages, allows a stream to automatically replay from the last successfully processed message, in case of persistent failures. If set to true, it will always auto-commit (if auto-commit is enabled). If not set (default), it effectively has the same value as enableDlq, auto-committing erroneous messages if they are sent to a DLQ, and not committing them otherwise.
Default: not set.

I interpret this as, if my `@StreamListener` runs into an error, it will retry up to `max-attempts` then write to DLQ.
Once the record is sent to DLQ, the offset should be committed. However, when I check the consumer group offset `$ kafka-consumer-groups --bootstrap-server 127.0.0.1:9092 --group my-consumer-group --describe`, the current offset is not incremented.

This is my application configuration
```
spring:
  cloud:
    stream:
      default:
        group: my-consumer-group
      kafka:
        binder:
          brokers:
            - 127.0.0.1:9092
          producer-properties:
            retries: 2147483647
        bindings:
          input:
            consumer:
              auto-commit-offset: false
              autoCommitOnError: true # by default it it same as enableDlq
              enableDlq: true
              dlqName: test-dlq
              configuration:
                max.poll.interval.ms: 5000
      bindings:
        input:
          destination: test-topic
        output:
          destination: test-topic
```

My StreamListener is similar to below:
I made an error case on purpose to see how DLQ and acks are handled.
```
@EnableBinding(SinkBinding.class)
public class KafkaConsumer {

    @StreamListener(value = SinkBinding.INPUT)
    public void listen(Message<String> message) throws InterruptedException {
        System.out.println("log this message : " + message.getPayload() + " offset : " + message.getHeaders().get(KafkaHeaders.OFFSET).toString());

        String a = null;
        a.substring(0, 3); // causing null pointer exception for testing purposes
        sendAck(message);
    }

    public void sendAck(Message<String> message) {
        Acknowledgment acknowledgment = message.getHeaders().get(KafkaHeaders.ACKNOWLEDGMENT, Acknowledgment.class);
        if (acknowledgment != null)  {
            try {
                acknowledgment.acknowledge();
                System.out.println("manual commit complete");
            }
[...]

PR #871, opened 2020-03-24
Offset commit when DLQ is enabled and manual ack

sobychacko:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/870

When an error occurs, if the application uses manual acknowldegment
(i.e. autoCommitOffset is false) and DLQ is enabled, then after
publishing to DLQ, the offset is not committed currently.
Addressing this issue by manually commiting after publishing to DLQ.

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -102,6 +102,7 @@ import org.springframework.kafka.listener.ConsumerAwareRebalanceListener;
 import org.springframework.kafka.listener.ConsumerProperties;
 import org.springframework.kafka.listener.ContainerProperties;
 import org.springframework.kafka.listener.DefaultAfterRollbackProcessor;
+import org.springframework.kafka.support.Acknowledgment;
 import org.springframework.kafka.support.KafkaHeaderMapper;
 import org.springframework.kafka.support.KafkaHeaders;
 import org.springframework.kafka.support.ProducerListener;
@@ -112,6 +113,7 @@ import org.springframework.kafka.support.converter.MessagingMessageConverter;
 import org.springframework.kafka.transaction.KafkaAwareTransactionManager;
 import org.springframework.kafka.transaction.KafkaTransactionManager;
 import org.springframework.lang.Nullable;
+import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.messaging.MessageHeaders;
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -1483,6 +1502,10 @@ public void onSuccess(SendResult<K, V> result) {
 							KafkaMessageChannelBinder.this.logger
 									.debug("Sent to DLQ " + sb.toString());
 						}
+						if (!DlqSender.this.properties.getExtension().isAutoCommitOffset()) {
```
garyrussell: Unfortunately, we need to get the actual `AckMode` from the listener container - the user might have reconfigured it to MANUAL_IMMEDIATE via the container customizer.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -1483,6 +1506,10 @@ public void onSuccess(SendResult<K, V> result) {
 							KafkaMessageChannelBinder.this.logger
 									.debug("Sent to DLQ " + sb.toString());
 						}
+						if (ackMode == ContainerProperties.AckMode.MANUAL || ackMode == ContainerProperties.AckMode.MANUAL_IMMEDIATE) {
+							final Acknowledgment acknowledgment = messageHeaders.get(KafkaHeaders.ACKNOWLEDGMENT, Acknowledgment.class);
+							acknowledgment.acknowledge();
```
garyrussell: No need for a variable here.

```suggestion
							messageHeaders.get(KafkaHeaders.ACKNOWLEDGMENT, Acknowledgment.class).acknowledge();
```

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -1446,14 +1466,17 @@ boolean isTopicPattern() {
 	private final class DlqSender<K, V> {
 
 		private final KafkaTemplate<K, V> kafkaTemplate;
+		private final ExtendedConsumerProperties<KafkaConsumerProperties> properties;
 
-		DlqSender(KafkaTemplate<K, V> kafkaTemplate) {
+		DlqSender(KafkaTemplate<K, V> kafkaTemplate, ExtendedConsumerProperties<KafkaConsumerProperties> properties) {
```
garyrussell: You don't need the properties any more.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -214,6 +216,8 @@
 
 	private KafkaExtendedBindingProperties extendedBindingProperties = new KafkaExtendedBindingProperties();
 
+	private Map<ConsumerDestination, ContainerProperties.AckMode> ackModeInfo = new HashMap<>();
```
garyrussell: Probably safer to use a `ConcurrentHashMap<>()` we're not iterating so shouldn't be a problem but, just in case...

PR #878, opened 2020-04-03
Refactor binder provisioning producer factory destroy right away.

mhyeon-lee:
When channel binding, destroy the producerFactory after provisioning. (none transaction producer)
Often blocking a lot of time when joining `ioThread` to close KafkaProducer.
PhysicalCloseTimeout for forced shutdown without waiting for time Set the value to 1.

I think there is no problem because it is a temporary ProducerFactory for provisioning.

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -351,11 +351,11 @@ public class KafkaMessageChannelBinder extends
 		 */
 		KafkaAwareTransactionManager<byte[], byte[]> transMan = transactionManager(
 				producerProperties.getExtension().getTransactionManager());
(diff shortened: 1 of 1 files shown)


garyrussell:
It's not clear what you are trying to achieve here.

    // force close right away after provisioning

The timeout is passed to the `close()` operation - it's how long to wait for it to complete; it has nothing to do with when the close() is issued.

Please explain what's driving this change.

mhyeon-lee:
@garyrussell
I checked again. It was my wrong setting.

I didn't set `transaction-id-prefix` because I'm not using transactional producer.
But I had `enable.idempotence: true` set, so KafkaProducer set TransactionManager.
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L519

Then, when close is executed, it is blocked in `ioThread.join(timeout)` by communication with TransactionManager.  (https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L1204)

In conclusion, there was no blocking delay by removing the `enable.idempotence` setting.

Sorry for the confusion.

I'll close this PR.

Thank you.

PR #904, opened 2020-05-21
GH-899: num.stream.threads not taking effect

sobychacko:
Binder is overriding the num.stream.threads property specified
through the binder configuration. Fixing this issue.

Adding tests to verify.

Doc changes

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/899


garyrussell:
@sobychacko This seems incredibly complicated. Why is it so much more complex than simply falling back to the binder-level property when there is no binding-level property?

sobychacko:
Issue created to tackle the architecture of how we handle concurrency in the binder: https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/905

Issue #935, opened 2020-07-12
KafkaNullConverter should support MimeTypeUtils.ALL


garyrussell:
It currently supports no `MimeType`s.

Converter is not invoked by `NegotiatingMessageConverterWrapper` in Spring Cloud Function.

Also see https://github.com/spring-cloud/spring-cloud-function/issues/557

PR #936, opened 2020-07-13
GH-935: Fix KafkaNullConverter supported MimeType

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/935

Spring Cloud Function now checks if a converter supports the mime type before
calling it. Previously, the converter supported no mime types, so it was never
called, breaking Kafka Tombstone record processing (outbound).

The converter must support all mime types so it can perform a no-op conversion,
retaining the `KafkaNull`.

The abstract converter will return null whenever the payload is not a `KafkaNull`,
which is a signal to spring-cloud-function to try the next converter.


sabbyanandan:
Addressed in https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/83c42a55e9d32d85fa3c58ca41b5353477ab215f.

PR #953, opened 2020-08-22
fix health indicator to properly indicate partition failure

CleverChuk:
topicInformation#partitionInfos never actually contains the partitionInfo retrieved from the brokers therefore the original code will only return health down when all brokers are down and the request times out. With this update, health down will be returned if any of the partitions do not have a leader.

### DIFF
--- a/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
+++ b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaBinderConfigurationProperties.java
@@ -16,21 +16,10 @@
 
 package org.springframework.cloud.stream.binder.kafka.properties;
 
-import java.time.Duration;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
-import javax.validation.constraints.AssertTrue;
-import javax.validation.constraints.Min;
-import javax.validation.constraints.NotNull;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.producer.ProducerConfig;
-
 import org.springframework.boot.autoconfigure.kafka.KafkaProperties;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.stream.binder.HeaderMode;
@@ -40,6 +29,15 @@ import org.springframework.expression.Expression;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
+
+import javax.validation.constraints.AssertTrue;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import java.time.Duration;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Configuration properties for the Kafka binder. The properties in this class are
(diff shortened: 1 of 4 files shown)


sobychacko:
@CleverChuk Could you please clarify what you meant by `topicInformation#partitionInfos never actually contains the partitionInfo retrieved from the broker`?

CleverChuk:
> @CleverChuk Could you please clarify what you meant by `topicInformation#partitionInfos never actually contains the partitionInfo retrieved from the broker`?

Yes, 
line 135 in the source when the function (topicInformation.getPartitionInfos()) is called the partitionInfo list returned will never contain any of the one returned in line 132. In essence, we have a logical error there. On the other hand, if that logic was to pass, the second conditional will throw a NPE if the partition has no leader.

sobychacko:
@CleverChuk I guess my question was more towards finding out why it never returned the info from `topicInformation.getPartitionInfos()`. The [binder class](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java) sets that information on startup. Wonder if there is something wrong there. Did you debug to find that out? Not saying that there isn't an issue, but I am curious. Regardless, your second point about NPE is very valid. Please comment on what you think about the first concern and we can hopefully merge this PR soon.

CleverChuk:
@sobychacko 
That's the root cause of the problem. The start up information is stale. So when it does the check with the current data from the brokers it won't find a match. Also, PartitionInfo doesn't override equals method which is called in contains hence in the case that the partitionInfo has the same data their reference will be different and it'll still fail.

sobychacko:
@CleverChuk Sorry for going back and forth, but we just want to make sure that this change is right. The problem I see is that, when we make this call `this.metadataConsumer.partitionsFor(topic)`, it will return all the partitions for the topic regardless of the current consumer's partitions. Now imagine that you have 2 consumers running in the same consumer group and a topic with multiple partitions. With the changes in this PR, it will check for partitions that this consumer may not be consuming at the moment. The check here for `topicInformation.getPartitionInfos().contains(partitionInfo)` ensures that we are only taking into consideration the current consumer's partitions. Now, the point about `equals()` is a valid concern which we can address. I still want to find out why the information goes stale. Did you identify the issue there?

CleverChuk:
@sobychacko Totally understand. It goes stale because the `partitionInfos` that was gathered at startup doesn't reflect what is returned by `this.metadataConsumer.partitionsFor(topic)` hence when a partition loses a leader regardless of whether it's being consumer by this consumer or not it wouldn't tell you until all the partitions loses their leader. At that point the whole cluster is down then it gives you a timeout. Also look at this [comment](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/f2e543c816187e6d5116b0c293c679118119da1a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L958).

garyrussell:
I think Soby's point is valid.

While I agree that if there is no leader for a partition, the binder can be considered down for publishing, but only if this instance actually publishes to that partition.

It is, however, possible to get a false down state if this instance only ever publishes to partition 0 (e.g. using its instance index); it is healthy even if partition 1 has no leader (and this binder has no consumers on partition 1). Furthermore, there may be no producer bindings at all (e.g. a Sink app). On the consumer side, we should definitely only consider the assigned partitions.

Perhaps a false DOWN is better than the current false UP due to the `contains()` problem but I don't think the binder can make blind assumptions about the apps that are using it. Maybe we need a new configuration flag - `considerDownWhenAnyPartitionHasNoLeader`.

sobychacko:
@garyrussell Thanks for the feedback! 
@CleverChuk I like the idea proposed by Gary above, i.e introduce a new flag `considerDownWhenAnyPartitionHasNoLeader` and if the user enables this, then the change introduced by this PR gets activated. What do you think about that?

CleverChuk:
@sobychacko 
I like it too.

sobychacko:
You can add it as a binder level property in `KafkaBinderConfigurationProperties`.

sobychacko:
This PR is now merged upstream. Cherry picked the changes to `3.0.x` branch. 
@CleverChuk Thanks for this PR! I added some minor changes on top of your commits and added some docs. Squashed all the commits into a single commit before pushing upstream. Looking forward to more PR contributions from you!

Issue #952, opened 2020-08-18
KafkaMessageChannelBinder sets retries to 0


marbon87:
In https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/5b3b92cdb9237e8a64707d4bd1395e9d97e90551/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L495 
retries for the producer is set to 0.
The default from kafka is MAX_INT (https://docs.confluent.io/current/installation/configuration/producer-configs.html#retries)

If there is no good reason to override the default from kafka, the configuration in KafkaMessageChannelBinder should be removed.

garyrussell:
https://gitter.im/spring-cloud/spring-cloud-stream?at=5f3be18a750a27413029c0ca

garyrussell:
The `BUFFER_MEMORY_CONFIG` setting looks suspicious too - it is currently the same as the default https://kafka.apache.org/documentation/#buffer.memory and unnecessary.

PR #955, opened 2020-08-26
Remove unnecessary configs from binder

sobychacko:
Following two properties are removed from the producer configs in the binder.

(ProducerConfig.RETRIES_CONFIG, 0)
(ProducerConfig.BUFFER_MEMORY_CONFIG, 33554432)

In the case of RETRIES, this is a bug to override the default in Kafka client (MAX_INT) and
in the latter case, this is unnecessary as this is the same default in the client.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/952


garyrussell:
Cherry-picked to 3.0.x as a57531ce

Issue #808, opened 2019-11-26
Enhancement: Support Coercing `replication-factor` when default and Broker defines a policy


derrick-anderson:
Use Case:
When using the spring cloud stream binder, I would like to be able to auto-provision topics with minimal boilerplate code. When connecting to a broker that defines a policy for items like `replication-factor` i would like the binder to coerce the default selection (in this case 1) to the required selection (in this case 3). 

Reproduction:
When using Confluent Cloud as my Kafka provider, I can connect to the broker successfully by attaching my `configuration` set to the `spring.cloud.stream.kafka.binder.configuration`.  However, if spring attempts to create a Topic that does not exist the following error will be thrown.

```
ERROR 19109 --- [           main] o.s.c.s.b.k.p.KafkaTopicProvisioner      : Failed to create topics
org.apache.kafka.common.errors.PolicyViolationException: Topic replication factor must be 3 
```
Preventing topic creation and ultimately application startup.

Solutions:
I see that the Kafka Topic provisioner does have a feature currently that coerces the partition tolerance down if allowed by the configuration properties and notifies the user as such.  Can we include a default override for policy violations such as this?  For example, knowing that I set the value as 2 means that i would see an error but leaving the default as 1 would allow the system to coerce it to the broker defined policy value of 3.

Thanks for reading, I look forward to input.
-Derrick

sobychacko:
@derrick-solstice Have you tried setting the replication factor as a binding level topic property? `spring.cloud.stream.kafa.bindings.input.consumer.topic.replicationFactor` (change `input` to the actual binding name) will be used by the provisioner. If you set that to `3`, then I think the binder will create the topic with a replication factor of `3`.

derrick-solstice:
@sobychacko Thanks for the response,  yes I was able to modify my `replication-factor` property on the binder.  

My enhancement proposal was more toward providing a method for respecting policies at the broker level.  Particularly when they just change the default setting and have no functionality impact.  The reason i believe there is precedent here is that it's already supported for partitions through a tolerate flag.

In this particular scenario, if Confluent was to change their broker policy after i've deployed DataFlow, I would suddenly stop connecting as no new topics would be provisioned.

Further, this was originally encountered while implementing Spring Cloud Data Flow on GKE and it is quite hard to change the service map properties after deployment without a full re-creation.

sobychacko:
[...]

PR #956, opened 2020-08-27
Change default replication factor to -1

sobychacko:
Binder now uses a default value of -1 for replication factor signaling the
broker to use defaults. Users who are on Kafka brokers older than 2.4,
need to set this to the previous default value of 1 used in the binder.

In either case, if there is an admin policy that requires replication factor > 1,
then that value must be used instead.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/808

### DIFF
--- a/docs/src/main/asciidoc/overview.adoc
+++ b/docs/src/main/asciidoc/overview.adoc
@@ -106,9 +106,9 @@ spring.cloud.stream.kafka.binder.replicationFactor::
 The replication factor of auto-created topics if `autoCreateTopics` is active.
 Can be overridden on each binding.
 +
-NOTE: If you are using Kafka broker versions prior to 2.4, then this value should be set to `1`.
-Starting with broker version 2.4, the binder will use `-1` as the default value, which indicates the broker to use the same default set on the broker.
-Check with your Kafka broker admins to see if there is a policy in place that requires the replication factor to be greater than 1, in which case, that value should be used instead of any defaults used by the binder.
+NOTE: If you are using Kafka broker versions prior to 2.4, then this value should be set to at least `1`.
+Starting with version 3.0.8, the binder uses `-1` as the default value, which indicates that the broker 'default.replication.factor' property will be used to determine the number of replicas.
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/overview.adoc
```diff
@@ -106,7 +106,11 @@ spring.cloud.stream.kafka.binder.replicationFactor::
 The replication factor of auto-created topics if `autoCreateTopics` is active.
 Can be overridden on each binding.
 +
-Default: `1`.
+NOTE: If you are using Kafka broker versions prior to 2.4, then this value should be set to `1`.
```
garyrussell, line 109:
```suggestion
NOTE: If you are using Kafka broker versions prior to 2.4, then this value should be set to at least `1`.
```

docs/src/main/asciidoc/overview.adoc
```diff
@@ -106,7 +106,11 @@ spring.cloud.stream.kafka.binder.replicationFactor::
...
+Starting with broker version 2.4, the binder will use `-1` as the default value, which indicates the broker to use the same default set on the broker.
```
garyrussell, line 110:
```suggestion
Starting with broker version 3.1, the binder uses `-1` as the default value, which indicates that the broker 'default.replication.factor' property will be used to determine the number of replicas.
```

docs/src/main/asciidoc/overview.adoc
```diff
@@ -106,7 +106,11 @@ spring.cloud.stream.kafka.binder.replicationFactor::
...
+Check with your Kafka broker admins to see if there is a policy in place that requires the replication factor to be greater than 1, in which case, that value should be used instead of any defaults used by the binder.
```
garyrussell, line 111:
```suggestion
Check with your Kafka broker admins to see if there is a policy in place that requires a minimum replication factor, if that's the case then, typically, the `default.replication.factor` will match that value and `-1` should be used, unless you need a replication factor greater than the minimum.
```

garyrussell:
And cherry-picked to 3.0.x

Issue #877, opened 2020-04-02
Deprecate/Rename autoCommitOffset - much confusion with enable.auto.commit


jskim1991:
Looking at binder reference

> autoCommitOffset
The following properties are available for Kafka consumers only and must be prefixed with spring.cloud.stream.kafka.bindings.channelName.consumer..
...
Default: true.

Looking at spring-kafka reference
> Starting with version 2.3, the framework sets enable.auto.commit to false unless explicitly set in the configuration. Previously, the Kafka default (true) was used if the property was not set.

When using Spring Boot 2.2.5.RELEASE and Spring Cloud Hoxton.SR3, somehow `enable.auto.commit` seems to be always false

![스크린샷 2020-04-02 오후 11 27 40](https://user-images.githubusercontent.com/17305081/78261472-3e2e5c00-753a-11ea-98d4-65f7b78c0a90.png)
![스크린샷 2020-04-02 오후 11 26 06](https://user-images.githubusercontent.com/17305081/78261475-3ff81f80-753a-11ea-9508-b1ddd0810475.png)

garyrussell:
They are different properties, unfortunately.

`autoCommitOffset` is a binder property that controls the container's ack mode.

`auto.commit.offset` is a Kafka consumer property which we always disable by default because we prefer to reliably manage the offsets by the container (or use code) rather than have the kafka-clients commit them in an indeterminate manner.

The name of the binder property is unfortunate.

jskim1991:
@garyrussell due to the name, I have always thought that `autoCommitOffset` was a binder property which controls the Kafka consumer property `enable.auto.commit`. Thanks for the clarification.

garyrussell:
Sorry, I had a typo in my answer.

I understand the confusion.

I think we should deprecate `autoCommitOffset` and allow direct control of the container `ackMode` via a property.

@sobychacko WDYT?

sobychacko:
@garyrussell That sounds like a good idea to me. The binder property `autoCommitOffset` is always a source of confusion as it implies a Kafka property. We can deprecate that use `ackMode` as you suggested. Reopening the issue to track that. Thanks.

PR #957, opened 2020-09-03
Deprecate autoCommitOffset in favor of ackMode

sobychacko:
Deprecate autoCommitOffset in favor of using a newly introduced consumer property ackMode.
If the consumer is not in batch mode and if ackEachRecord is enabled, then container
will use RECORD ackMode. Otherwise, use the provided ackMode using this property.
If none of these are true, then it will defer to the default setting of BATCH ackMode
set by the container.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/877

### DIFF
--- a/docs/src/main/asciidoc/overview.adoc
+++ b/docs/src/main/asciidoc/overview.adoc
@@ -175,6 +175,8 @@ By default, offsets are committed after all records in the batch of records retu
 The number of records returned by a poll can be controlled with the `max.poll.records` Kafka property, which is set through the consumer `configuration` property.
 Setting this to `true` may cause a degradation in performance, but doing so reduces the likelihood of redelivered records when a failure occurs.
 Also, see the binder `requiredAcks` property, which also affects the performance of committing offsets.
+This property is deprecated as of 3.1 in favor of using `ackMode`.
+If the `ackMode` is not set and batch mode is not enabled, `RECORD` ackMode will be used.
 +
 Default: `false`.
 autoCommitOffset::
--- a/docs/src/main/asciidoc/overview.adoc
+++ b/docs/src/main/asciidoc/overview.adoc
@@ -183,9 +183,14 @@ If set to `false`, a header with the key `kafka_acknowledgment` of the type `org
 Applications may use this header for acknowledging messages.
 See the examples section for details.
 When this property is set to `false`, Kafka binder sets the ack mode to `org.springframework.kafka.listener.AbstractMessageListenerContainer.AckMode.MANUAL` and the application is responsible for acknowledging records.
-Also see `ackEachRecord`.
+Also see `ackEachRecord`. This property is deprecated as of 3.1. See `ackMode` for more details.
 +
 Default: `true`.
+ackMode::
+Specify the container ack mode.
+This is based on the AckMode enumeration defined in Spring Kafka.
+If `ackEachRecord` property is set to `true` and consumer is not in batch mode, then this will use the ack mode of `RECORD`, otherwise, use the provided ack mode using this property.
+
 autoCommitOnError::
 Effective only if `autoCommitOffset` is set to `true`.
 If set to `false`, it suppresses auto-commits for messages that result in errors and commits only for successful messages. It allows a stream to automatically replay from the last successfully processed message, in case of persistent failures.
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/overview.adoc
```diff
@@ -183,9 +183,14 @@ If set to `false`, a header with the key `kafka_acknowledgment` of the type `org
...
 See the examples section for details.
 When this property is set to `false`, Kafka binder sets the ack mode to `org.springframework.kafka.listener.AbstractMessageListenerContainer.AckMode.MANUAL` and the application is responsible for acknowledging records.
-Also see `ackEachRecord`.
+Also see `ackEachRecord`. This property is deprecated as of 3.1. See `ackMode` for more details.
 +
 Default: `true`.
+ackMode::
+Specify the container ack mode.
+This is based on the AckMode enumeration defined in Spring Kafka.
+If `ackEachRecord` property is set to `true` and consumer is not in batch mode, then this will use the ack mode of `RECORD`, otherwise, use the provided ack mode using this property.
```
garyrussell, line 192: Why not deprecate `ackEachRecord` too?

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -655,17 +655,11 @@ else if (getApplicationContext() != null) {
...
-			messageListenerContainer.getContainerProperties()
-					.setAckMode(ContainerProperties.AckMode.MANUAL);
-			messageListenerContainer.getContainerProperties().setAckOnError(false);
-		}
-		else {
-			messageListenerContainer.getContainerProperties()
-					.setAckOnError(isAutoCommitOnError(extendedConsumerProperties));
-			if (extendedConsumerProperties.getExtension().isAckEachRecord()) {
+		final ContainerProperties.AckMode ackMode = extendedConsumerProperties.getExtension().getAckMode();
+		if (!extendedConsumerProperties.isBatchMode()) { //if batch mode, we will ignore any ack mode.
```
garyrussell, line 659: We need to allow ack modes other than RECORD with batch mode; it's just that RECORD mode won't work with batchMode so that should be ignored.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -655,17 +655,11 @@ else if (getApplicationContext() != null) {
...
-		if (!extendedConsumerProperties.getExtension().isAutoCommitOffset()) {
-			messageListenerContainer.getContainerProperties()
-					.setAckMode(ContainerProperties.AckMode.MANUAL);
-			messageListenerContainer.getContainerProperties().setAckOnError(false);
-		}
-		else {
-			messageListenerContainer.getContainerProperties()
-					.setAckOnError(isAutoCommitOnError(extendedConsumerProperties));
-			if (extendedConsumerProperties.getExtension().isAckEachRecord()) {
+		final ContainerProperties.AckMode ackMode = extendedConsumerProperties.getExtension().getAckMode();
```
garyrussell, line 658: ? We've only deprecated `ackEachRecord` not disabled it.

I suggest

```java
if (ackMode == null && ...isAckEachRecord()) {
    ackMode = AckMode.RECORD;
}
```

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -1283,6 +1277,7 @@ else if (record == null) {
 									((MessagingException) message.getPayload())
 											.getFailedMessage());
 					if (ack != null) {
+
```
garyrussell, line 1280: ?

garyrussell (changes requested):

Issue #809, opened 2019-11-26
KafkaBinderMetrics.computeUnconsumedMessages blocks prometheus actuator


robwruck:
Currently, the metric `spring.cloud.stream.binder.kafka.offset` is implemented as a Gauge. That means that every request for `/actuator/prometheus` triggers a synchronous call to `computeUnconsumedMessages` which in turn waits for up to 60 seconds (DEFAULT_TIMEOUT) for a result from Kafka.
A Gauge is not meant to run lengthy or asynchronous computations. When a Kafka broker becomes temporarily unavailable, this Gauge will cause `/actuator/prometheus` to respond with a 60 seconds delay, which might be too much for the Prometheus scraper resulting in **no metrics** being recorded until the broker is available again.
Since you are already spawning a new ThreadPoolExecutor on every call to `computeUnconsumedMessages`, it would be more efficient to use a single ThreadPoolExecutor to periodically calculate this metric in the background and use the Gauge to retrieve the most recent value.

sobychacko:
@robwruck Thank you for the suggestion. That sounds like a reasonable approach to me. You are most welcomed to contribute a PR with these enhancements which we can then review!

PR #965, opened 2020-09-28
Kafka binder metrics improvements

sobychacko:
KafkaBinderMetrics has a blocking call in which it waits for the default timeout of
60 seconds if Kafka broker is down. This happens for each topic within a consumer
group. Refactor this code, so that we have this check performed in a periodic task
and if the runtime check fails to return within a smaller timewindow (5 seconds),
return immediately by providing the latest value from the periodic task results.

Periodic task for computing the lags is run every 60 seconds.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/809

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
@@ -25,6 +25,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
@@ -62,7 +63,9 @@ import org.springframework.util.ObjectUtils;
 public class KafkaBinderMetrics
 		implements MeterBinder, ApplicationListener<BindingCreatedEvent> {
 
-	private static final int DEFAULT_TIMEOUT = 60;
+	private static final int DEFAULT_TIMEOUT = 5;
+
+	private static final int DELAY_BETWEEN_TASK_EXECUTION = 60;
 
 	private static final Log LOG = LogFactory.getLog(KafkaBinderMetrics.class);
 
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -200,7 +230,6 @@ private long computeUnconsumedMessages(String topic, String group) {
 	@Override
 	public void onApplicationEvent(BindingCreatedEvent event) {
 		if (this.meterRegistry != null) {
-			// meters are idempotent when called with the same arguments so safe to call
 			// it multiple times
```
garyrussell, line 233: ? the remaining comment makes no sense.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -114,45 +124,38 @@ public void bindTo(MeterRegistry registry) {
...
-							.entrySet()) {
-						OffsetAndMetadata current = metadataConsumer
-								.committed(endOffset.getKey());
-						lag += endOffset.getValue();
-						if (current != null) {
-							lag -= current.offset();
-						}
-					}
-				}
+				lag = findAnyLag(topic, group, lag, this.metadataConsumers);
```
garyrussell, line 158: Maybe change to `findTotalTopicGroupLag`?

`lag` doesn't need to be a parameter.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -114,45 +124,38 @@ public void bindTo(MeterRegistry registry) {
...
+				this.unconsumedMessages.put(topic + "-" + group, lag);
+			}
+			catch (Exception ex) {
+				LOG.debug("Cannot generate metric for topic: " + topic, ex);
+			}
+		};
+	}
+
+	private long computeAndGetUnconsumedMessages(String topic, String group) {
 		ExecutorService exec = Executors.newSingleThreadExecutor();
```
garyrussell, line 153: Creating and destroying an executor for each topic/group seems a little excessive. Can't we use a thread pool executor?

garyrussell (changes requested):

garyrussell (approved):

PR #975, opened 2020-10-15
Change Log etc. Replication Factor

garyrussell:
If the user has not explicitly set the `SreamsConfig.REPLICATION_FACTOR_CONFIG`,
set it from the binder property.

This is used for infrastructure topics (change logs and repartition topics).


sobychacko:
Looks good. Merged upstream.

Issue #966, opened 2020-10-01
Support supplying custom destination resolver when sending to DLQ


sdobberstein:
With the way the "sendToDlq" works today, you have two options:
1) Specify a DLQ name and that is used.
2) Don't specify a DLQ name and instead it defaults to ```error.<topic>.<group>```

It would be nice if there was a way to supply your own custom destination resolver instead.

**Example use cases:**  
1) I would like to be able to route to different topics based on the type of exception.  Serialization errors to one topic for example, and networking/code errors to another topic.
2) I would like to be able to route to different topics based on values within the headers.  For example, I might store the retry count as a header.  Until you hit a certain number of retries route it to "topic a", but once you hit some threshold then route it to "topic b".

sdobberstein:
Here is a link to the relevant code where the destination resolver is currently being defined:
https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/7bebe9b78fd5c53aba45e1160c29932a5216de30/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderUtils.java#L105-L126

sobychacko:
@sdobberstein Are you asking for customizing the DLQ name when the binder tries to resolve the destination for the DLQ based on some criteria in the `ConsumerRecord` or the exception? That is the impression I am getting from the request. If we resort to this, then how the partition function is applied is completely up to the user-provided destination resolver function. Also, in the Kafka Streams binder, the only error that we support in the context of sending to DLQ is any deserialization errors. Any other type of errors such as logic related must be addressed in the application explicitly and automatic DLQ handling won't work there. With that said, this sounds like a reasonable enhancement to add to the binder. Are you interested in contributing a patch for this as a PR?

PR #976, opened 2020-10-20
Custom DLQ Destination Resolver

sobychacko:
Allow applications to provide a custom DLQ destination resolver
implementaiton by providing a new interface DlqDestinationResolver
as part of binder's public contract. This interface is a BiFunction
extension using which the applications can provide more fine grained
control over where to route records in error.

Adding test to verify.

Adding docs.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/966

### DIFF
--- a/docs/src/main/asciidoc/kafka-streams.adoc
+++ b/docs/src/main/asciidoc/kafka-streams.adoc
@@ -832,13 +832,41 @@ When the above property is set, all the records in deserialization error are aut
 
 You can set the topic name where the DLQ messages are published as below.
 
+You can provide an implementation for `DlqDestinationResolver` which is a functional interface.
+`DlqDestinationResolver` takes `ConsumerRecord` and the exception as inputs and then allows to specify a topic name as the output.
+By gaining access to the Kafka `ConsumerRecord`, the header records can be introspected in the implementation of the `BiFunction`.
+
+Here is an example of providing an implementation for `DlqDestinationResolver`.
+
+[source]
+----
+@Bean
+public DlqDestinationResolver dlqDestinationResolver() {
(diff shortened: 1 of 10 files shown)


docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -832,13 +832,41 @@ When the above property is set, all the records in deserialization error are aut
...
+        }
+        else {
+            return "topic2-dlq";
+        }
+    };
+}
+----
+
+One important thing to keep in mind when providing an implementation for `DlqDestinationResolver` is that the provisioner in the binder will not auto create topics for the application.
+This is because there is no way for the binder to infer all the DLQ topics the implementation is creating or their names.
```
garyrussell, line 857:
```suggestion
This is because there is no way for the binder to infer the names of all the DLQ topics the implementation might send to.
```

garyrussell (changes requested):
We need a similar change in the message channel binder, for consistency.

docs/src/main/asciidoc/kafka-streams.adoc
sobychacko: @garyrussell Do you recommend the message channel binder changes for this in this PR or as a separate one?

garyrussell: I personally would do it in the same PR, but it's up to you.

garyrussell:
Merged, but I don't think it will work with a named binder - see https://github.com/spring-cloud/spring-cloud-stream/issues/1541

garyrussell:
Forget that.

Issue #972, opened 2020-10-13
Kafka key is not populated when using kafka-binding with Flux


sabbyanandan:
@rgonciarz commented on [Thu Oct 08 2020](https://github.com/spring-cloud/spring-cloud-stream/issues/2030)

**Describe the issue**
Kafka key is not populated when using kafka-binding with Flux

```
    @Bean
    fun topologyR() = Function<Flux<String>,Flux<String>> {
        it.doOnNext { v -> log.info("topologyR: {}", v) }
    }
```

Setup:
topic1 -> topologyA (kafka streams binding) -> topic2 
topic2 -> topologyR (kafka flux binding) -> topic3
topic3 -> topologyB (kafka streams binding)

**To Reproduce**
I created a small project that may help to reproduce the issue:
https://github.com/rgonciarz/reactive-topology-chain/tree/reactive-function-no-avro](https://github.com/rgonciarz/reactive-topology-chain/tree/reactive-function-no-avro)

1. Run application, and either send a message on topic1 (String,String serialization) or use ApplicationKtTest 

Result:
```
2020-10-08 19:20:02.221  INFO 14232 --- [-StreamThread-1] wr.reactive.TopologyConfig               : topologyA: 9307f81e-7148-459e-a49a-e5aeb07cacac, 9307f81e-7148-459e-a49a-e5aeb07cacac
2020-10-08 19:20:02.324  INFO 14232 --- [container-0-C-1] wr.reactive.TopologyConfig               : topologyR: 9307f81e-7148-459e-a49a-e5aeb07cacac
2020-10-08 19:20:02.326  INFO 14232 --- [-StreamThread-1] wr.reactive.TopologyConfig               : topologyB: null, 9307f81e-7148-459e-a49a-e5aeb07cacac
```
We loose information about Kafka key. To workaround the problem we need to use Message abstraction and manually copy header from KafkaHeaders.RECEIVED_MESSAGE_KEY to KafkaHeaders.MESSAGE_KEY.

Affected version: Spring boot 2.3.3

---

@garyrussell commented on [Mon Oct 12 2020](https://github.com/spring-cloud/spring-cloud-stream/issues/2030#issuecomment-707227465)

The framework cannot make any assumption that the inbound key should be propagated to the outbound key.

Perhaps an option could be added, but your work around is the correct way to handle this currently.

garyrussell:
>Perhaps an option could be added, but your work around is the correct way to handle this currently.

[...]

PR #978, opened 2020-10-26
Expand messageKeyExpression docs

sobychacko:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/972

### DIFF
--- a/docs/src/main/asciidoc/overview.adoc
+++ b/docs/src/main/asciidoc/overview.adoc
@@ -360,6 +360,11 @@ messageKeyExpression::
 A SpEL expression evaluated against the outgoing message used to populate the key of the produced Kafka message -- for example, `headers['myKey']`.
 With versions before 3.0, the payload could not be used unless native encoding was being used because, by the time this expression was evaluated, the payload was already in the form of a `byte[]`.
 Now, the expression is evaluated before the payload is converted.
+In the case of a regular processor (`Function<String, String>` or `Function<Message<?>, Message<?>`), if the produced key needs to be same as the incoming key from the topic, this property can be set as below.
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/overview.adoc
```diff
@@ -360,6 +360,11 @@ messageKeyExpression::
 A SpEL expression evaluated against the outgoing message used to populate the key of the produced Kafka message -- for example, `headers['myKey']`.
 With versions before 3.0, the payload could not be used unless native encoding was being used because, by the time this expression was evaluated, the payload was already in the form of a `byte[]`.
 Now, the expression is evaluated before the payload is converted.
+In the case of a regular processor (`Function<String, String>` or `Function<Message<?>, Message<?>`), if the produced key needs to be same as the incoming key from the topic, this property can be set as below.
+`spring.cloud.stream.kafka.bindings.<output-binding-name>.producer.messageKeyExpression: headers['kafka_receivedMessageKey']`
+There is an important caveat to keep in mind for reactive functions.
+In that case, it is up to the application to manually copy the headers from the incoming messages to outbound messages.
+Otherwise, this property will not have any effect on reactive functions.
```
garyrussell, line 367: After further review, I think we should reword this to

```suggestion
You can set the header, e.g. `myKey` and use `headers['myKey']` as suggested above or, for convenience, simply set the `KafkaHeaders.MESSAGE_KEY` header, and you do not need to set this property at all.
```

WDYT?

sobychacko: Looks good! Added another commit with this change.

Issue #945, opened 2020-08-03
Retry and exponential backoff configuration in the latest release.


puneetbabbar26:
Hi,

I am trying to get retry with exponential backoff working with my Spring cloud Kafka streams app which uses the latest plugins (HOXTON SR6)

The following don't seem to have any impact on the behavior of the application when the exception (e.g RestClientException) is propagated from the application to the listener.

      consumer:
        autoCommitOnError: false
        max-attempts: 2147483647
        backOffInitialInterval: 1000
        backOffMaxInterval: 300000
        backOffMutliplier: 2.0

I saw these configurations being suggested at a few places like -
https://stackoverflow.com/questions/51182670/exactly-once-delivery-is-it-possible-through-spring-cloud-stream-binder-kafka-or

and they have been documented in the previous releases of the framework.
https://docs.spring.io/spring-kafka/docs/2.4.0.RELEASE/reference/html/#stateful-retry

Is there a way to do it via configuration with the current implementation? spring-cloud-stream-binder-kafka-streams:3.0.6

garyrussell:
I am not sure why your properties are not being applied, but retrying indefinitely in the binder like that will cause serious problems with Kafka - it will likely cause a rebalance.

Instead, you should disable retry in the binder and configure a `SeekToCurrentErrorHandler` with infinite retries - that way Kafka will know the consumer is still active. See this answer on Stack Overflow... https://stackoverflow.com/questions/63193500/replay-kafka-topic-from-the-last-successful-message/63194089#63194089

puneetbabbar26:
Thanks for getting back to me Gary. 

The initial configuration I posted was just to get it to work, I don't intend to keep the retries to be more than 5.
The problem is to get it to work.
I tried what you suggested and the configuration is still not getting applied and the Exceptions are not getting handled by the ErrorHandler and the message retried.

Getting the following exception

```
Exception in thread "DataEnricher-enrichCOHeader-applicationId-7106172b-5eda-44a3-9411-46e66e603bec-StreamThread-1" java.lang.IllegalStateException: some error
	at a.c.r.c.o.l.d.f.stream.DataEnricher$1.transform(DataEnricher.java:119)
	at a.c.r.c.o.l.d.f..stream.DataEnricher$1.transform(DataEnricher.java:106)
	at org.apache.kafka.streams.kstream.internals.TransformerSupplierAdapter$1.transform(TransformerSupplierAdapter.java:47)
[...]

PR #980, opened 2020-10-27
Allow retries in Kafka Streams binder

sobychacko:
Provide applications the capability to retry critical sections of the
business logic. This is accomplished through a new API using which
critical path can be wrapped inside a Callable.

Adding tests and docs.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/945


docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -926,6 +926,84 @@ This implies that if there are multiple functions or `StreamListener` methods in
 Unlike the support for deserialization exception handlers as described above, the binder does not provide such first class mechanisms for handling production exceptions.
 However, you still can configure production exception handlers using the `StreamsBuilderFactoryBean` customizer which you can find more details about, in a subsequent section below.
 
+=== Retrying critical business logic
+
+There are scenarios in which you might want to retry parts of your business logic that are critical to the application.
+There maybe an external call to a relational database or invoking a REST endpoint from the Kafka Streams processor.
+These calls can fail for various reasons such as network issues or remote service unavailability.
+More often, these failures may self resolve if you can try them again.
+Kafka Streams binder provides an API that allows you to wrap the critical sections of the business logic in retrying temlate.
```
garyrussell, line 935:
```suggestion
The Kafka Streams binder provides an API that allows you to wrap the critical sections of the business logic in retrying template.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -926,6 +926,84 @@ This implies that if there are multiple functions or `StreamListener` methods in
...
+                }
+
+                @Override
+                public void close() {
+                }
+            });
+}
+```
+
+As this is a termainl operation, it is recommended to use this API with low level Processor API in Kafka Streams.
```
garyrussell, line 968:
```suggestion
As this is a terminal operation, it is recommended to use this API with the low level Processor API in Kafka Streams.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -926,6 +926,84 @@ This implies that if there are multiple functions or `StreamListener` methods in
...
+
+Here is another example of using a different API from this class, but this time it allows you to use it on non-termainal operation.
```
garyrussell, line 970:
```suggestion
Here is another example of using a different API from this class, but this time it allows you to use it on non-terminal operation.
```

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/CriticalPathRetryingKafkaStreamsDelegate.java
```diff
@@ -0,0 +1,186 @@
...
+				}
+			});
+		}
+		catch (Exception e) {
+			return null;
+		}
+	}
+
+	private RetryTemplate getRetryTemplate() {
+		RetryTemplate retryTemplate = new RetryTemplate();
```
garyrussell, line 172: We should't create a new template for each operation.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/CriticalPathRetryingKafkaStreamsDelegate.java
```diff
@@ -0,0 +1,186 @@
...
+ * ({@link org.apache.kafka.streams.kstream.Transformer} operations to be wrapped inside the {@link Callable} implementation.
+ * <p>
+ * See the relevant API methods for more information.
+ *
+ * @param <K> Key type
+ * @param <V> Value type
+ * @author Soby Chacko
+ * @since 3.0.9
+ */
+public class CriticalPathRetryingKafkaStreamsDelegate<K, V> {
```
garyrussell, line 54: It's not at all clear what this simple wrapper offers when you can use a `RetryTemplate` directly, with a much richer API - different retry policies, different back-off policies, retry context, recovery strategies to be called when retries are exhausted.

garyrussell (changes requested):
Sorry, Soby, I don't like this at all; I think it would be better to call `AbstractBinder.buildRetryTemplate()` and register it as a bean for use by the binding. That would use the properties that the original issue poster was concerned about.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/CriticalPathRetryingKafkaStreamsDelegate.java
sobychacko: I was just thinking of this as a starting point. We can add the capability to send to some sort of DLT in the future if the users want to send the failing k/v for later retrieval, but continue processing in the main processor.

sobychacko:
@garyrussell , isn't the configs through `ConsumerProperties` only applicable per input binding? These properties are specific at the binder level and not tied with the consumer retry properties. But, if we can reuse those and use it as you suggested, we can certainly go with that.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/CriticalPathRetryingKafkaStreamsDelegate.java
sobychacko: I have been recommending the usage of the `RetryTemplate` directly before, but somehow it never caught on as a usage pattern.

garyrussell:
Oh, sorry, yes; you are right; that is a binder-level setting. But we could still register that as a bean and even use it in your wrapper, perhaps (or the user can use it directly if (s)he wants to use the richer API). If you make it injectable as well (in the wrapper), and they want different properties at the binding level, they can create their own template and inject it.

garyrussell:
No; these are binding consumer properties not binder.

garyrussell:
I was thinking we could register a bean `<bindingName>RetryTemplate`

garyrussell:
No; these are binding consumer properties not binder; sorry; I was confused by the package name - looks like that class should be in binding, not binder.

I was thinking we could register a bean <bindingName>RetryTemplate.

sobychacko:
if we create `<bindingName>RetryTemplate`, then by the time the application hits the critical section, it may lose its context. What if it went through some transformations and merges and then want to retry some operations? What `RetryTemplate` we will use in that scenario? If there is only a single input binding, to begin with, then that is okay. However, if we have more, the situation may become ambiguous as to which template to choose from.

garyrussell:
>it may lose its context. 

Sorry; I don't understand your concern. They know which binding the stream is bound to. Maybe we can zoom tomorrow? - I am busy with releases today and have to leave early.

sobychacko:
Per further discussions with @garyrussell we decided to rework the PR.  Removed the custom API that was added in the initial commit and now reusing the `RetryTemplate` based on consumer binding.

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -926,6 +926,97 @@ This implies that if there are multiple functions or `StreamListener` methods in
...
+By default, Kafka Streams binder creates `RetryTemplate` beans for all the input bindings.
+
+If the function has the following signature,
+```
+@Bean
+public java.util.function.Consumer<KStream<Object, String>> process()
+```
+and with default binding name, the `RetryTemplate` will be registered as `process-in-0-RetryTemplate`.
+This is following the convention of binding name (`process-in-0`) followed by the literal `-RetryTemplate`.
+In the case of multiple input bindings, there will be a separate `RetrtyTempate` bean available per binding.
```
garyrussell, line 944: Typo `RetrtyTempate`

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -926,6 +926,97 @@ This implies that if there are multiple functions or `StreamListener` methods in
...
+If there is a custom `RetryTemplate` bean is available in the application and provided through `spring.cloud.stream.bindings.<binding-name>.consumer.retryTemplateName`, then that takes precedence over any input binding level retry templates.
```
garyrussell, line 945:
```suggestion
If there is a custom `RetryTemplate` bean available in the application and provided through `spring.cloud.stream.bindings.<binding-name>.consumer.retryTemplateName`, then that takes precedence over any input binding level retry template configuration properties.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -926,6 +926,97 @@ This implies that if there are multiple functions or `StreamListener` methods in
...
+
+Once the `RetryTemplate` from the binding is injected in the application, that can be used to retry any critical sections of the application.
```
garyrussell, line 947:
```suggestion
Once the `RetryTemplate` from the binding is injected into the application, it can be used to retry any critical sections of the application.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -926,6 +926,97 @@ This implies that if there are multiple functions or `StreamListener` methods in
...
+                    }
+
+                    @Override
+                    public void close() {
+                    }
+                });
+    }
+}
+```
+
```
garyrussell, line 1019: Note that when retries are exhausted, by default, the last exception will be thrown, causing the `KStream` to terminate.
If you wish to handle the exception and continue processing, you can add a `RecoveryCallback` to the execute method:

```
retryTemplate.execute(context -> {
                       //Critical business logic goes here.
                    }, context -> {
                       //Recovery logic goes here.
                       return null;
                    ));
```

Refer to the [spring-retry project](https://github.com/spring-projects/spring-retry) for more information about the `RetryTemplate`, retry policies, backoff policies and more.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderUtils.java
```diff
@@ -131,6 +138,16 @@ static void prepareConsumerBinding(String name, String group,
 						kafkaStreamsBinderDlqRecoverer);
 			}
 		}
+
+		if (StringUtils.isEmpty(properties.getRetryTemplateName())) {
```
garyrussell, line 142: `isEmpty()` is deprecated in Spring 5.3; use `StringUtils.hasText()` instead.

garyrussell (changes requested):

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -926,6 +926,110 @@ This implies that if there are multiple functions or `StreamListener` methods in
...
+By default, Kafka Streams binder creates `RetryTemplate` beans for all the input bindings.
+
+If the function has the following signature,
+```
+@Bean
+public java.util.function.Consumer<KStream<Object, String>> process()
+```
+and with default binding name, the `RetryTemplate` will be registered as `process-in-0-RetryTemplate`.
+This is following the convention of binding name (`process-in-0`) followed by the literal `-RetryTemplate`.
+In the case of multiple input bindings, there will be a separate `RetrtyTemplate` bean available per binding.
```
garyrussell, line 944: Typo still there `RetrtyTemplate`.

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -926,6 +926,110 @@ This implies that if there are multiple functions or `StreamListener` methods in
...
+If there is a custom `RetryTemplate` bean available in the application and provided through `spring.cloud.stream.bindings.<binding-name>.consumer.retryTemplateName`, then that takes precedence over any input binding level retry template configuration properties.
+
+OOnce the `RetryTemplate` from the binding is injected into the application, it can be used to retry any critical sections of the application.
```
garyrussell, line 947: New typo `OOnce`.

garyrussell (changes requested):

Toomy:
Good Afternoon
Thank you for fixing this problem. As I have the same issue ( need to retry a message in the streams processor api) do you know when the fix will be availabe?

sobychacko:
@Toomy These changes are going to be part of the `3.0.9.RELEASE` which is scheduled to occur tomorrow (11/06/2020). Thanks!

PR #988, opened 2020-11-11
Fix documentation for kafka streams concurrency on binder level

### LINKED ISSUE #987: Wrong documentation for kafka streams concurrency on binder level
The current documented property `spring.cloud.stream.binder.configuration.num.stream.threads` does not work.

The correct one is `spring.cloud.stream.kafka.streams.binder.configuration.num.stream.threads` which is already covered by tests at MultipleFunctionsInSameAppTests#125

relates to #899 #844

gustavomonarin:
The property spring.cloud.stream.binder.configuration.num.stream.threads does not work and is silently ignored.

The property spring.cloud.stream.kafka.streams.binder.configuration.num.stream.threads works fine and  is already covered by tests at MultipleFunctionsInSameAppTests#125

resolves #987


sobychacko:
@gustavomonarin Thank you for the PR! Merging upstream.

PR #991, opened 2020-11-17
Do not perform alter topics call if no changes required

tdanylchuk:
There are a few purposes of this change request:
- omit the redundant call to the broker on application startup (decrease startup time)
- do not fail startup if the broker does not support INCREMENTAL_ALTER_CONFIGS
- remove a bit exception logs

So Recently I've tried to bump the cloud version to the latest and failed on deploying my app.
In tests I have broker with all features enabled whereas on dev environment I have features like INCREMENTAL_ALTER_CONFIGS disabled.

Application yet successfully started but stream never went up due to 
```
org.springframework.cloud.stream.provisioning.ProvisioningException: Provisioning exception; nested exception is java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.UnsupportedVersionException: The broker does not support INCREMENTAL_ALTER_CONFIGS 	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.createTopic(KafkaTopicProvisioner.java:339) 	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.doProvisionConsumerDestination(KafkaTopicProvisioner.java:224) 	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.provisionConsumerDestination(KafkaTopicProvisioner.java:190) 	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.provisionConsumerDestination(KafkaTopicProvisioner.java:86) 	at org.springframework.cloud.stream.binder.AbstractMessageChannelBinder.doBindConsumer(AbstractMessageChannelBinder.java:403) 	at org.springframework.cloud.stream.binder.AbstractMessageChannelBinder.doBindConsumer(AbstractMessageChannelBinder.java:91) 	at org.springframework.cloud.stream.binder.AbstractBinder.bindConsumer(AbstractBinder.java:143) 	at org.springframework.cloud.stream.binding.BindingService.lambda$rescheduleConsumerBinding$1(BindingService.java:201) 	at org.springframework.cloud.sleuth.instrument.async.TraceRunnable.run(TraceRunnable.java:68) 	at org.springframework.scheduling.support.DelegatingErrorHandlingRunnable.run(DelegatingErrorHandlingRunnable.java:54) 	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264) 	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304) 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) 	at java.base/java.lang.Thread.run(Thread.java:834) Caused by: java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.UnsupportedVersionException: The broker does not support INCREMENTAL_ALTER_CONFIGS 	at org.apache.kafka.common.internals.KafkaFutureImpl.wrapAndThrow(KafkaFutureImpl.java:45) 	at org.apache.kafka.common.internals.KafkaFutureImpl.access$000(KafkaFutureImpl.java:32) 	at org.apache.kafka.common.internals.KafkaFutureImpl$SingleWaiter.await(KafkaFutureImpl.java:104) 	at org.apache.kafka.common.internals.KafkaFutureImpl.get(KafkaFutureImpl.java:272) 	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.createTopicAndPartitions(KafkaTopicProvisioner.java:404) 	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.createTopicIfNecessary(KafkaTopicProvisioner.java:349) 	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.createTopic(KafkaTopicProvisioner.java:326) 	... 15 more Caused by: org.apache.kafka.common.errors.UnsupportedVersionException: The broker does not support INCREMENTAL_ALTER_CONFIGS
--
```

From this exception is not clear which bonder is actually causing an issue, so only debugging help.

Also, I have a question from this case: isn't healthcheck for binder supposed to be DOWN is this case?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -147,6 +147,11 @@ If set to `false`, the binder relies on the partition size of the topic being al
 If the partition count of the target topic is smaller than the expected value, the binder fails to start.
 +
 Default: `false`.
+spring.cloud.stream.kafka.binder.autoAlterTopics::
+If set to `true`, the binder alters destination topic configs if required.
+If set to `false`, the binder relies on existing configs of the topic.
++
+Default: `false`.
 spring.cloud.stream.kafka.binder.transaction.transactionIdPrefix::
 Enables transactions in the binder. See `transaction.id` in the Kafka documentation and https://docs.spring.io/spring-kafka/reference/html/_reference.html#transactions[Transactions] in the `spring-kafka` documentation.
 When transactions are enabled, individual `producer` properties are ignored and all producers use the `spring.cloud.stream.kafka.binder.transaction.producer.*` properties.
--- a/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
+++ b/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
@@ -169,7 +169,7 @@ public class KafkaTopicProvisioner implements
 						topicDescriptions.putAll(all.get(this.operationTimeout, TimeUnit.SECONDS));
 					}
 					catch (Exception ex) {
-						throw new ProvisioningException("Problems encountered with partitions finding", ex);
+						throw new ProvisioningException("Problems encountered with partitions finding for: " + name, ex);
 					}
 					return null;
 				});
@@ -242,7 +242,7 @@ public class KafkaTopicProvisioner implements
 					}
 				}
 				catch (Exception ex) {
-					throw new ProvisioningException("provisioning exception", ex);
+					throw new ProvisioningException("Provisioning exception encountered for " + name, ex);
 				}
 			}
 		}
@@ -312,7 +312,7 @@ public class KafkaTopicProvisioner implements
 					throw (Error) throwable;
 				}
 				else {
-					throw new ProvisioningException("provisioning exception", throwable);
+					throw new ProvisioningException("Provisioning exception encountered for " + name, throwable);
 				}
 			}
 			return new KafkaConsumerDestination(name, partitions, dlqTopic);
@@ -336,7 +336,7 @@ public class KafkaTopicProvisioner implements
 			else {
 				// TODO:
 				// https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/514#discussion_r241075940
-				throw new ProvisioningException("Provisioning exception", throwable);
+				throw new ProvisioningException("Provisioning exception encountered for " + name, throwable);
 			}
 		}
 	}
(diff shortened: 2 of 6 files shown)


spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
```diff
@@ -376,32 +376,7 @@ private void createTopicAndPartitions(AdminClient adminClient, final String topi
 		if (names.contains(topicName)) {
 			//check if topic.properties are different from Topic Configuration in Kafka
 			if (this.configurationProperties.isAutoCreateTopics()) {
```
garyrussell, line 378: I think this should be under a new property `autoAlterTopics` - false by default.

It will still break for users on old brokers where the topics have been changed without reflecting that change in the yaml; previously this would just be ignored.

garyrussell (commented):
@sobychacko I will leave it up to you as to whether you want to merge this as-is and follow up with another PR for my suggested new property.

spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
tdanylchuk: don't mind to add in the current PR, it was my initial plan, but just to make this PR simple kept with regular check

sobychacko: @tdanylchuk Please go ahead and add the property in this PR itself if you don't mind.

tdanylchuk: done, please review

sobychacko:
@tdanylchuk Thanks for the PR. It is now merged upstream and backported to `3.0.x`. Looking forward to more PR contributions.

README.adoc
```diff
@@ -147,6 +147,11 @@ If set to `false`, the binder relies on the partition size of the topic being al
 If the partition count of the target topic is smaller than the expected value, the binder fails to start.
 +
 Default: `false`.
+spring.cloud.stream.kafka.binder.autoAlterTopics::
+If set to `true`, the binder alters destination topic configs if required.
```
garyrussell, line 151: We should probably add a comment that the broker version must be at least 2.3 to use this feature.

tdanylchuk:
Guys @sobychacko @garyrussell sorry for rising this topic once again but i do interested in 
`Also, I have a question from this case: isn't healthcheck for binder supposed to be DOWN is this case?`
can healthcheck check kafkaConsumerContainer status also not the only topic and broker connectivity?

tdanylchuk:
Sorry for tagging you guys @sobychacko @garyrussell but I just want to clean the dust from this topic regarding my question 
in the previous post, we faced a couple of times, that service was unable to consume records and was healthy so we needed to spend some time on the investigation since we have alerts only for service health status.

sobychacko:
@tdanylchuk Could you start a new issue on this matter? It is easy to miss updates when commenting on closed issues. I didn't really fully understand the context, so it might be better to start a new thread/issue.

Issue #990, opened 2020-11-16
Support allowNonTransactional config for ProducerConfigurationMessageHandler 


mhyeon-lee:
Producers for batch processing may not need transactions.
Could you provide allowNonTransactional config to ProducerConfigurationMessageHandler and KafkaTemplate?

garyrussell:
This is a reasonable request.

In the meantime, there are two workarounds:

1. Use the multi-binder support to configure two binders; one transactional and one not.

2. Use a customizer with a bit of reflection:

```java
@SpringBootApplication
public class Kbgh990Application {

	public static void main(String[] args) {
		SpringApplication.run(Kbgh990Application.class, args);
	}

	@Autowired
	StreamBridge bridge;

	@Bean
	ApplicationRunner runner() {
		return args -> this.bridge.send("notTx", "foo");
	}

	@Bean
	ProducerMessageHandlerCustomizer<?> customizer() {
		return (handler, destinationName) -> {
			if (destinationName.equals("non.tx.topic")) {
				((KafkaProducerMessageHandler<?, ?>) handler).getKafkaTemplate().setAllowNonTransactional(true);
				new DirectFieldAccessor(handler).setPropertyValue("allowNonTransactional", true);
			}
		};
	}

}
```
```properties
spring.cloud.stream.kafka.binder.transaction.transaction-id-prefix=tx-
spring.cloud.stream.kafka.binder.required-acks=all
spring.cloud.stream.bindings.notTx.destination=non.tx.topic
```

mhyeon-lee:
@garyrussell 
Thank you for your consideration
I will apply the workaround you gave as a guide, and will switch when options are added later.

PR #999, opened 2020-12-03
Support allowNonTransactional config

sobychacko:
Add a new producer extended property for allowNonTransactional which when
set to true, tansactions will be disabled on the underlying KafkaTemplate.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/990

### DIFF
--- a/docs/src/main/asciidoc/overview.adoc
+++ b/docs/src/main/asciidoc/overview.adoc
@@ -443,6 +443,13 @@ Timeout in number of seconds to wait for when closing the producer.
 +
 Default: `30`
 
+allowNonTransactional::
+Normally, all output bindings associated with a transactional binder will publish in a new transaction, if one is not already in process.
+This property allows you to override that behavior.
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/overview.adoc
```diff
@@ -443,6 +443,11 @@ Timeout in number of seconds to wait for when closing the producer.
 +
 Default: `30`
 
+allowNonTransactional::
+If set to true, transactions will be disabled on this producer binding.
```
garyrussell, line 447:
```suggestion
Normally, all output bindings associated with a transactional binder will publish in a new transaction, if one is not already in process.
This property allows you to override that behavior.
If set to true, records published to this output binding will not be run in a transaction, unless one is already in process.
```

garyrussell (changes requested):
Just doc polishing; the commit comment is a bit misleading too, so you might want to force push a change to that too.

garyrussell (approved):

PR #1009, opened 2020-12-16
Fix topic name confg in docs

mle-enso:


sobychacko:
Thanks for the PR. Merged upstream.

PR #1010, opened 2020-12-29
fix property name in consuming batches

50qbits:
@pivotal-issuemaster This is an Obvious Fix

Change in property name


sobychacko:
Thanks for the PR. Merged upstream.

Issue #1014, opened 2021-01-18
unable to customize AdminClientConfig programmatically


colaho:
Due to security reason, our kafka credentials are stored in aws secrets manager and are extracted on the fly during the application bootstrap. With the `ProducerConfigCustomizer` and `ConsumerConfigCustomizer` beans, we can connect to our cluster successfully by injecting the SASL config into the producer / consumer properties. However, we are getting this error in the log.

> 2021-01-15 17:51:52.660  INFO 1 --- [| adminclient-6] o.a.k.c.a.i.AdminMetadataManager         : [AdminClient clientId=adminclient-6] Metadata update failed
> org.apache.kafka.common.errors.TimeoutException: Timed out waiting to send the call.

From [`KafkaTopicProvisioner`](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java). it seems that the provisioner always creates the AdminClient from properties. Is there any way which we could customize the AdminClientConfig via a `AdminClientConfigCustomizer` bean? Thanks

sobychacko:
Hi @colaho Here is a sample that demonstrates how to access a Kaka cluster with security by providing binder level properties. By providing it at the binder level, these security properties are applicable to all the components such as producer, consumer and admin client. 
https://github.com/spring-cloud/spring-cloud-stream-samples/tree/master/kafka-security-samples/kafka-ssl-demo

This is the config for this project: https://github.com/spring-cloud/spring-cloud-stream-samples/blob/master/kafka-security-samples/kafka-ssl-demo/src/main/resources/application.yml

Here is another project that uses multi-binder support with security: https://github.com/spring-cloud/spring-cloud-stream-samples/blob/master/multi-binder-samples/kafka-multi-binder-jaas/src/main/resources/application.yml

Please let us know if they satisfy your use case. Thanks!

colaho:
Hi @sobychacko, our case is a bit different from the examples. 

Because the credentials are secretly stored on the aws (in another words, we won't even know it in production) and are rotated regularly, we could not put it statically in the properties. As mentioned, for consumer and producer, we used the `ProducerConfigCustomizer` and `ConsumerConfigCustomizer` beans to dynamically inject the credentials and it works.

```java
    @Bean
[...]

PR #1023, opened 2021-02-02
AdminClient customizer

sobychacko:
Provide the ability for applications to customize AdminClient by
introducing a new interface AdminClientConfigCustomizer.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1014

### DIFF
--- a/docs/src/main/asciidoc/overview.adoc
+++ b/docs/src/main/asciidoc/overview.adoc
@@ -786,3 +786,20 @@ For example, if you want to gain access to a bean that is defined at the applica
 When the binder discovers that these customizers are available as beans, it will invoke the `configure` method right before creating the consumer and producer factories.
 
 Both of these interfaces also provide access to both the binding and destination names so that they can be accessed while customizing producer and consumer properties.
+
+[[admin-client-config-customization]]
+=== Customizing AdminClient Configuration
+
+As with consumer and producer config customization above, applications can also customize the configuration for admin client by providing an AdminClientConfigCustomizer.
+AdminClientConfigCustomizer's configure method provides access to the admin client properties, using which you can define further customization.
+Binder's Kafka topic provisioner gives the highest precedence for the properties given through this customizer.
+Here is an example of providing this customizer bean.
+
+```
+@Bean
+public AdminClientConfigCustomizer adminClientConfigCustomizer() {
+    return props -> {
+        props.put(CommonClientConfigs.SECURITY_PROTOCOL_CONFIG, "SASL_SSL");
+    };
+}
+```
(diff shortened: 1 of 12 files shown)


docs/src/main/asciidoc/overview.adoc
```diff
@@ -786,3 +786,20 @@ For example, if you want to gain access to a bean that is defined at the applica
 When the binder discovers that these customizers are available as beans, it will invoke the `configure` method right before creating the consumer and producer factories.
 
 Both of these interfaces also provide access to both the binding and destination names so that they can be accessed while customizing producer and consumer properties.
+
+[[admin-client-config-customization]]
+=== Customizing AdminClient Configuration
+
+As with consumer and producer config customization above, applications can also customize the configuration for admin client by providing an AdminClientConfigCustomizer.
```
garyrussell, line 793:
```suggestion
As with consumer and producer config customization above, applications can also customize the configuration for admin clients by providing an `AdminClientConfigCustomizer`.
```

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/AutoCreateTopicDisabledTests.java
```diff
@@ -97,7 +100,7 @@ public void testAutoCreateTopicDisabledFailsOnProducerIfTopicNonExistentOnBroker
 		configurationProperties.getConfiguration().put("max.block.ms", "3000");
 
 		KafkaTopicProvisioner provisioningProvider = new KafkaTopicProvisioner(
-				configurationProperties, kafkaProperties);
+				configurationProperties, kafkaProperties, adminClientConfigCustomizer);
```
garyrussell, line 103: It's not clear why we're changing this test since we aren't verifying that the customizer was called. It's already covered by an earlier test.

Ditto the next test.

garyrussell (changes requested):
I wonder if we should add the binding name to these customizer interfaces and customize a new set of properties for each binding?

Perhaps another issue, though, for a future release.

Another issue is how can they tell which binder is being customized when named binders are being used. I suppose they can use the broker addresses to determine it, but perhaps it would be nice to pass them the binder name. Again, new feature, perhaps.

docs/src/main/asciidoc/overview.adoc
sobychacko: Good points about the bindings/binder names. I will add separate issues for those.

garyrussell: I meant all the customizers (producer, consumer) not just the admin.

garyrussell (approved):

PR #1039, opened 2021-03-11
Remove binder interface supported in Spring Kafka

sobychacko:
StreamsListener (for Micrometer) is now directly available
in Spring Kafka starting from 2.5.3 (as KafkaStreamsMicrometerListener).
Removing the temporory interface added in the binder.


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/StreamsBuilderFactoryManager.java
```diff
@@ -82,9 +78,6 @@ public synchronized void start() {
 				for (StreamsBuilderFactoryBean streamsBuilderFactoryBean : streamsBuilderFactoryBeans) {
 					streamsBuilderFactoryBean.start();
 					this.kafkaStreamsRegistry.registerKafkaStreams(streamsBuilderFactoryBean);
-					if (this.listener != null) {
-						this.listener.streamsAdded("streams." + n++, streamsBuilderFactoryBean.getKafkaStreams());
-					}
```
garyrussell, line 87: ?? We still need to detect and/or add a `KafkaStreamsMicrometerListener` and tell it about streams added and removed.

Don't we have any tests that ensure the meters are registered properly? If we do, I would expect them to fail.

garyrussell (changes requested):

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/StreamsBuilderFactoryManager.java
sobychacko: oops, that is completely an oversight on my part. I will correct this.

sobychacko: @garyrussell I had to create a new PR branch for this. Closing this one in favor of that one.

Issue #850, opened 2020-02-19
Why doesn't KStreamBinder have a lifecycle for DefaultBinding?


jaehun2841:
I previously asked stackoverflow.
(https://stackoverflow.com/questions/60282225/why-doesnt-kstreambinder-have-a-lifecycle-for-defaultbinding)

I'm using KStream parameter in @StreamBuilder.
This will create a DefaultBinding through KStreamBinder.
My requirement is to use Binding visualization and control.

However KStreamBinder.java creates DefaultBinding as below.
```java
return new DefaultBinding <> (name, null, outboundBindTarget, null);
```

You can't control the state via springboot /actuator/bindings because the lifecycle is null.
`(POST /actuator/bindings/{bindings-name} {"state":"PAUSED"})`
How can I control the state of a Binding?

**Are you not intentionally providing flow control for KafkaStreams?**

The version I am using is below.
* org.springframework.cloud:spring-cloud-stream:3.0.1.RELEASE
* org.springframework.cloud:spring-cloud-stream-binder-kafka-core:3.0.1.RELEASE
* org.springframework.cloud:spring-cloud-stream-binder-kafka-streams:3.0.1.RELEASE
* org.apache.kafka:kafka-streams:2.3.1

please answer about my question. 

Let me show you the code

```kotlin
@EnableBinding(Dispatcher::class)
class MessageDispatcher {
  @StreamListener(Dispatcher.INBOX)
  override fun dispatch(input: KStream<String, Message>) {
   dispatch(input)
  }
}
```
```kotlin
interface Dispatcher {
   companion object {
      const val INBOX = "inbox"
   }

  @Input(INBOX)
  fun input(): KStream<String, Message>
}
```

please discuss
Thank you.

jaehun2841:
Please review @sobychacko

sobychacko:
@jaehun2841 Sorry for the delay in responding to the issue. It is intentional that we don't support this feature in Kafka Streams binder in the same way it is supported in the regular Kafka binder. In the regular binder, we are delegating to Spring Kafka and there we are pausing/resuming the poller. In Kafka Streams, this consumer is not exposed. See this [comment](https://stackoverflow.com/questions/39048923/stop-a-kafka-streams-app#comment65453255_39048923). What exactly is your use case? Are you trying to pause and then resume a live stream processing system?

[...]

PR #1042, opened 2021-03-12
Support KStream lifecycle through binding endpoint

sobychacko:
Introduce the ability for Kafka Streams application's lifecycle
management through actuator binding endpoints. Kafka Streams
only supports STOP and START operations. PAUSE/RESUME operations
that is available in regular message channel based binders
are not available in Kafka Streams binder.

Adding tests and docs.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1038
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/850

https://stackoverflow.com/questions/60282225/why-doesnt-kstreambinder-have-a-lifecycle-for-defaultbinding

### DIFF
--- a/docs/src/main/asciidoc/kafka-streams.adoc
+++ b/docs/src/main/asciidoc/kafka-streams.adoc
@@ -1692,9 +1692,9 @@ public class KafkaStreamsApplication {
 
 As we can see, the application has two Kafka Streams functions - one, a consumer and another a function.
 The consumer binding is named by default as `consumer-in-0`.
-Similarly, for the function, the input binding is `function-in-0` and `function-out-0`.
+Similarly, for the function, the input binding is `function-in-0` and the output binding is `function-out-0`.
 
-Once the application started, we can find details about the bindings using the following bindings endpoint.
+Once the application is started, we can find details about the bindings using the following bindings endpoint.
 
 ```
  curl http://localhost:8080/actuator/bindings | jq .
--- a/docs/src/main/asciidoc/kafka-streams.adoc
+++ b/docs/src/main/asciidoc/kafka-streams.adoc
@@ -1640,6 +1640,118 @@ For instance, if we want to change the header key on this binding to `my_event` 
 
 `spring.cloud.stream.kafka.streams.bindings.process-in-0.consumer.eventTypeHeaderKey=my_event`.
 
+=== Binding visualization and control in Kafka Streams binder
+
+Starting with version 3.1.2, Kafka Streams binder supports binding visualization and control.
+The only two lifecycle phases supported are `STOPPED` and `STARTED`.
+The lifecycle phases `PAUSED` and `RESUMED` are not available in Kafka Streams binder.
+
+In order to activate binding visualization and control, the application needs to include the following two dependencies.
+
+```
+<dependency>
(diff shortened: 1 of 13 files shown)


docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1640,6 +1640,118 @@ For instance, if we want to change the header key on this binding to `my_event`
...
+	public Function<KStream<String, String>, KStream<String, String>> function() {
+		return ks -> ks;
+	}
+
+}
+```
+
+As we can see, the application has two Kafka Streams functions - one, a consumer and another a function.
+The consumer binding is named by default as `consumer-in-0`.
+Similarly, for the function, the input binding is `function-in-0` and `function-out-0`.
```
garyrussell, line 1695:
```suggestion
Similarly, for the function, the input binding is `function-in-0` and the output binding is `function-out-0`.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1640,6 +1640,118 @@ For instance, if we want to change the header key on this binding to `my_event`
...
+
+Once the application started, we can find details about the bindings using the following bindings endpoint.
```
garyrussell, line 1697:
```suggestion
Once the application is started, we can find details about the bindings using the following bindings endpoint.
```

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBindingInformationCatalogue.java
```diff
@@ -50,7 +50,7 @@
 
 	private final Map<KStream<?, ?>, KafkaStreamsConsumerProperties> consumerProperties = new ConcurrentHashMap<>();
 
-	private final Set<StreamsBuilderFactoryBean> streamsBuilderFactoryBeans = new HashSet<>();
+	private final Map<String, StreamsBuilderFactoryBean> streamsBuilderFactoryBeanPerBinding = new HashMap<>();
```
garyrussell, line 53: `dlqProducerFactories` needs to be changed to a map so you can `reset()` the PF (to close producer) when the binding is stopped.

garyrussell (changes requested):

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/GlobalKTableBinder.java
```diff
@@ -106,7 +106,7 @@ public synchronized void stop() {
 
 				final List<ProducerFactory<byte[], byte[]>> dlqProducerFactories =
 						kafkaStreamsBindingInformationCatalogue.getDlqProducerFactory(streamsBuilderFactoryBean);
```
garyrussell, line 108: This list is no longer used.

garyrussell (changes requested):

garyrussell (approved):

Issue #1046, opened 2021-03-18
Refactor DlqSender to Wait For the SendResultFuture to Complete


garyrussell:
The `SendResult` is being processed asynchronously, causing the offset commit to be queued and will incorrectly commit the wrong offset with MANUAL_IMMEDIATE AckMode.

See https://gitter.im/spring-cloud/spring-cloud-stream?at=6050f43595e23446e43cacd1

Temporary work around is to use MANUAL AckMode instead (so all commits are queued in order).

PR #1048, opened 2021-03-19
GH-1046: Fix Out of Order Offset Commit with DLQ

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1046

When using ack mode `MANUAL_IMMEDIATE`, we must acknowledge the delivery on
the container thread, otherwise the commit will be queued for later processing,
possibly causing out of order commits, incorrectly reducing the committed offset.

Wait for the send to complete before acknowledging; use a timeout slightly larger
than the configured producer delivery timeout to avoid premature timeouts.

Tested with user-provided test case.

**cherry-pick to 3.0.x**

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -1148,7 +1148,7 @@ public class KafkaMessageChannelBinder extends
 				sendTimeout = ((Number) timeout).longValue() + 2000L;
 			}
 			else if (timeout instanceof String) {
-				sendTimeout = Long.parseLong((String) timeout);
+				sendTimeout = Long.parseLong((String) timeout) + 2000L;
 			}
 			if (timeout == null) {
 				sendTimeout = ((Integer) ProducerConfig.configDef()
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -1141,8 +1142,21 @@ protected MessageHandler getErrorMessageHandler(final ConsumerDestination destin
 			final KafkaTemplate<?, ?> kafkaTemplate = new KafkaTemplate<>(
 					producerFactory);
 
+			Object timeout = producerFactory.getConfigurationProperties().get(ProducerConfig.DELIVERY_TIMEOUT_MS_CONFIG);
+			Long sendTimeout = null;
+			if (timeout instanceof Number) {
+				sendTimeout = ((Number) timeout).longValue() + 2000L;
+			}
+			else if (timeout instanceof String) {
+				sendTimeout = Long.parseLong((String) timeout);
```
sobychacko, line 1151: Why are we not increasing the timeout here as in the other cases?

sobychacko (changes requested):

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
garyrussell: We should; pushed.

sobychacko:
LGTM. Merged upstream and backported.

Issue #1051, opened 2021-03-24
When destinationIsPattern: true ,kafka stream bind ignore binding level Sede setting


xuxiankun:
``` yaml
spring:
   cloud:
     stream:
       bindings:
        process-in-0:
          destination: customer1.*
      kafka:
        streams:
          binder:
            configuration:
              metadata.max.age.ms: 30000
          bindings:
            process-in-0:
              consumer:   
                destination-is-pattern: true
                key-serde: org.apache.kafka.common.serialization.Serdes$StringSerde
                value-serde: org.apache.kafka.common.serialization.Serdes$StringSerde
``` 

	@Bean
    public Function<KStream<String, String>,KStream<String, String>> process() {
       return (input) ->{
        
        KStream kstream = input.map((key, value) -> {
                   
            return new KeyValue<>(key,value);
        });
        kstream.to("test");
        return kstream;
       };
i am getting exception like:

stream-thread [process-applicationId-2a4f4e1f-39ce-4e1d-834e-36b625c90553-StreamThread-1] Shutdown complete
Exception in thread "process-applicationId-2a4f4e1f-39ce-4e1d-834e-36b625c90553-StreamThread-1" org.apache.kafka.streams.errors.StreamsException: ClassCastException invoking Processor. Do the Processor's input types match the deserialized types? Check the Serde setup and change the default Serdes in StreamConfig or provide correct Serdes via method parameters. Make sure the Processor can accept the deserialized input of type key: [B, and value: [B.
Note that although incorrect Serdes are a common cause of error, the cast exception might have another cause (in user code, for example). For example, if a processor wires in a store, but casts the generics incorrectly, a class cast exception could be raised during processing, but the cause would not be wrong Serdes.
        at org.apache.kafka.streams.processor.internals.ProcessorNode.process(ProcessorNode.java:146)

sobychacko:
@xuxiankun Thanks for this report. It turned out to be a problem in the binder where we call the `stream` method on `StreamsBuilder. I submitted a [PR](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/1054) to address this issue. As a workaround, you can switch to the framework provided message conversion on the inbound thus skipping the inbound Serdes. Use the property: `spring.cloud.stream.bindings.<binding>.consumer.use-native-decoding: false`. This introduces some additional processors to your topology though. Next releases (`3.1.3` and `3.0.12.RELEASE`) will have this fix.

PR #1054, opened 2021-03-25
Fix destination as pattern issues in Kafka Streams

sobychacko:
When destination-is-pattern property is enabled and native deserialization
is used, then Kafka Streams binder does not use the correct Serde types.
Fixing this issue by providing the proper Serdes to the StreamsBuilder.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1051


garyrussell:
cherry-picked as f84876b9 after resolving conflict.

Issue #1081, opened 2021-05-21
Suspicious multiplication of ScheduledExecutorService in the method "bindTo" of class "KafkaBinderMetrics"


krocodl:
@Override
	public void bindTo(MeterRegistry registry) {
		this.scheduler = Executors.newScheduledThreadPool(this.binder.getTopicsInUse().size());
...
}

As a result, during the initialization of the application at some point, we can observe a lot of threads (for 33 topics, for example, more than 550), the maintenance of which slows down the process and sometimes even results in OOM

sobychacko:
@krocodl The thread pool, in this case, is upper bounded by the maximum number of topics. Therefore, you should only observe 33 threads at the most for 33 topics, I am not sure how you get to see 550. Did you do any further triaging to find out how it spawns that many threads?

krocodl:
First of all you created the pool with 1 thread. After you lost the reference on it and created the pool with 2 threads. But, the first pool is not yet collected by GC and now you have 3 threads together. And so on. Each thread does nothing, but it takes system memory and takes part in the scheduling process. After 30 topics, you potentially have (30+1)*15=465 threads. It is already the serious load on the switching contexts. 
In my case, the concurrency of the consumer was 5 and cpu was bounded by K8 settings. As a result these pools were not collected by GC in time and the instance was crashed with OOM. In the dump I found more than 2000 threads and start my investigations about strange count of threads. 
I think, that it is necessary to shutdown the existing pool before the new creation.

krocodl:
After that I started the service locally, waited for the end of the initialization and immediately made thread dump. In it I found all these threads.

sobychacko:
@krocodl Thank you for this analysis. We will try to fix this asap. Pull requests are welcomed to address this issue. If you plan to fix it, please let us know. Otherwise, we will do it on our end.

krocodl:
My next work day is Tuesday. I will try to make a patch that will subscribe to the same event and keep track of the fact that the scheduler field has changed, causing a cleanup for the old version of the value. This will allow me to fix the problem quickly and finally test my hypothesis. In that case, I will make a patch for binder as well. Thanks for the quick response!

sobychacko:
Sure, thanks for the update. We have a release scheduled for `3.1.3` next week. Hopefully, we can try to include it in the release.

krocodl:
Proof of the suspicions (the thread dump after start and how "pool-x-thread-y" like threads are created):

[...]

PR #1083, opened 2021-05-25
GH-1081 Suspicious multiplication of ScheduledExecutorService

krocodl:
GH-1081 Suspicious multiplication of ScheduledExecutorService in the method "bindTo" of class "KafkaBinderMetrics"

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1081

The wrong consequence of scheduler.shutdown() absence:
1) First of all we created the pool with 1 thread.
2) After we lost the reference on it and created the pool with 2 threads.
3) But, the first pool is not yet collected by GC and now you have 3 threads together. And so on.
Each thread does nothing, but it takes system memory and takes part in the scheduling process. After 30 topics, for example, we potentially have (30+1)*15=465 threads. It is already a serious additional load on the switching contexts and the native memory.

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
@@ -65,8 +65,6 @@ import org.springframework.util.ObjectUtils;
 public class KafkaBinderMetrics
         implements MeterBinder, ApplicationListener<BindingCreatedEvent> {
 
-    private static final int SCHEDULER_SHUTDOWN_TIMEOUT_SEC = 1;
-
     private static final int DEFAULT_TIMEOUT = 5;
 
     private static final int DELAY_BETWEEN_TASK_EXECUTION = 60;
@@ -126,15 +124,6 @@ public class KafkaBinderMetrics
         if (this.scheduler != null) {
             LOG.info("Try to shutdown the old scheduler with " + ((ScheduledThreadPoolExecutor) scheduler).getPoolSize() + " threads");
             this.scheduler.shutdown();
-            boolean terminationResult;
-            try {
-                terminationResult = this.scheduler.awaitTermination(SCHEDULER_SHUTDOWN_TIMEOUT_SEC, TimeUnit.SECONDS);
-            } catch (Exception ex) {
-                throw new IllegalStateException("The previous scheduler can't be terminated properly", ex);
-            }
-            if (!terminationResult) {
-                throw new IllegalStateException("The previous scheduler can't be terminated in " + SCHEDULER_SHUTDOWN_TIMEOUT_SEC + " seconds");
-            }
         }
 
         this.scheduler = Executors.newScheduledThreadPool(this.binder.getTopicsInUse().size());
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -114,33 +117,51 @@ public void setTimeout(int timeout) {
...
-	}
+    public void bindTo(MeterRegistry registry) {
+        /**
+         * We can't just replace one scheduler with another.
+         * Before and even after the old one is gathered by GC, it's threads still exist, consume memory and CPU resources to switch contexts.
+         * Theoretically, as a result of processing n topics, there will be about (1+n)*n/2 threads simultaneously at the same time.
+         */
+        if (this.scheduler != null) {
+            LOG.info("Try to shutdown the old scheduler with " + ((ScheduledThreadPoolExecutor) scheduler).getPoolSize() + " threads");
+            this.scheduler.shutdown();
```
garyrussell, line 128: Why not just `shutdownNow()`? The scheduler thread could be waiting in `consumer.endOffsets()`; not sure 1 second is long enough if you don't want to interrupt.

garyrussell (commented):
LGTM, just one comment.

garyrussell:
We should probably set a cap on the scheduler size - we really don't need a thread per topic; it won't hurt to queue some requests in the scheduler.

Maybe just use a single scheduler with, say, 5 threads? Then we can avoid all this shutdown/recreate stuff.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
krocodl: To be honest, I prefer not to interrupt code that is not completely under my control and that doesn't handle this signal in a very clear for me way. So I suggest just removing the wait. In my OOM case it is enough, that is, at least we will make it "not worse".

krocodl:
> Maybe just use a single scheduler with, say, 5 threads? Then we can avoid all this shutdown/recreate stuff.

First of all, I prefer to be conservative: for most projects, this code works well, and let's keep it as is. For me, it is not clear why the authors recreated the Scheduler instead of submitting one another task. One single thread pool is an obvious idea and I suspect they had some reasons to follow the more complex way. I am afraid of break something.

Secondly, in this case, we will make uncomparable already gathered and new metrics data. Especially, if gathering can take a long time, as we suspected in the previous discussion. 

In general, I would suggest that refactoring this code into a separate task to at least thoroughly compare the results before and after.

sobychacko:
@krocodl Thank you for this PR that cleaned up the `KafkaBinderMetrics` scheduling behavior. We can address any further enhancements in subsequent changes. I merged the PR upstream after some minor polishing.

krocodl:
@sobychacko Thanks you too!

sobychacko:
@krocodl More PR's are welcomed if you find any opportunities for enhancements!

krocodl:
@sobychacko we can think, about re-implementation of this code:
https://blog.devgenius.io/exceptions-and-retry-policy-in-kafka-6f0d3cf2c330?source=friends_link&sk=894c5caa71de80e331b0d78fdd3d88e9

PR #1115, opened 2021-08-04
Docs: rename property and fix typo

MitchelNijdam-Rockstars:
As far as I can tell the property value `logAndSkip` does not exist, instead `logAndContinue` is used.

### DIFF
--- a/docs/src/main/asciidoc/kafka-streams.adoc
+++ b/docs/src/main/asciidoc/kafka-streams.adoc
@@ -991,7 +991,7 @@ One important thing to keep in mind when providing an implementation for `DlqDes
 This is because there is no way for the binder to infer the names of all the DLQ topics the implementation might send to.
 Therefore, if you provide DLQ names using this strategy, it is the application's responsibility to ensure that those topics are created beforehand.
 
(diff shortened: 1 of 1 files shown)


MitchelNijdam-Rockstars:
The Contributer License Agreement page says: "If you are submitting an obvious fix, then you do not need to sign the CLA." so please let me know if it necessary in this case.

sobychacko:
@MitchelNijdam-Rockstars Thanks for this PR. There is [this value](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/main/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/DeserializationExceptionHandler.java#L45) out there which is wrongly documented as `logAndSkip`. It should be `skipAndContinue`. What it does is silently skip and continue without logging anything. We can update that in the docs. Feel free to close this PR. In the meantime, if you see further gaps in the docs/code, please send more PR's.

sobychacko:
@MitchelNijdam-Rockstars Changes are applied [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/commit/d7907bbdcc7196229668110c9ae7a38ad82efd15). Closing the PR.

Issue #1107, opened 2021-07-18
A lot of logs are produced due to metrics


rodion-fisenko:
Hi!

I found an issue due to which produces a lot of redundant logs like this:
`ConsumerCoordinator: [Consumer clientId=consumer-progress-pass-4, groupId=progress-pass] Found no committed offset for partition bingoblitz_app_progress_pass_steps_passed-1`

After some investigation, I found that the issue can be reproduced in the following conditions:
1. Use just initialised partition which has not received any messages yet. It this case the partition doesn't have committed offset;
2. Offset auto commit should be disabled (value of the property `spring.cloud.stream.kafka.binder.consumer-properties.enable.auto.commit` should be `false`);
3. Micrometer metrics are enabled (the bean `io.micrometer.core.instrument.MeterRegistry` exists in application context).

In this case the logs like described above will be produced every 60 seconds for each partition with no committed offset.

It happens because after merging of this PR (https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/965) some metrics collect every 60 seconds. 
As I see in the code the class `KafkaBinderMetrics` [create](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/54ac274ea32997221c5ca3c6c4ae7605345047a1/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java#L197) new consumer to calculate topic group lag. Offsets for this consumer cannot be auto reset to proper values (0 in this case) like it happens to usual consumers. 

Like a workaround to fix this issue we can use one of this options:
 1. Disable logs with level below `ERROR` for `org.apache.kafka.clients.consumer.internals.ConsumerCoordinator` logger;
2. Set the following property `spring.cloud.stream.kafka.binder.consumer-properties.enable.auto.commit: true` to enable auto commit. But this option [is not preferable](https://docs.spring.io/spring-kafka/reference/html/#committing-offsets), as it [can lead to problems](https://newrelic.com/blog/best-practices/kafka-consumer-config-auto-commit-data-loss).

Do you have a plan somehow to rework KafkaBinderMetrics to fix the issue? 

**UPD**
I found more suitable solution to avoid this issue. 

If we set the value of the property ContainerProperties.assignmentCommitOption to ALWAYS then uncommitted offsets be automatically committed during first partition assignment. 

To do it we should provide the following bean:

```
@Bean
[...]

PR #1123, opened 2021-08-18
GH-1107: Kafka metrics lag check

sobychacko:
When KafkaBinderMetrics checks lags for offsets in partitions, if the partitions
have nothing committed yet, it unnecessarily tries to find the last committed offset.
This is causing some unncessary logging. Avoid this by only checking the committed
offsets in a partition, if the current end offset is greater than zero.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1107

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 the original author or authors.
+ * Copyright 2016-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,10 +16,12 @@
 
 package org.springframework.cloud.stream.binder.kafka;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -210,11 +212,16 @@ private long findTotalTopicGroupLag(String topic, String group, Map<String, Cons
 
 		for (Map.Entry<TopicPartition, Long> endOffset : endOffsets
 				.entrySet()) {
```
garyrussell, line 214: Why not get all the committed offsets in a single call here `.committed(topicPartitions)` rather that one-at-a-time in the loop below?

garyrussell (changes requested):
I am not sure what's driving this change; when I tested this, I simply got offset=0 for partitions that had no committed offset.
```java
@KafkaListener(id = "kbgh1123", topics = "kbgh1123")
public void listen(String in, Consumer<?, ?> consumer) {
	Map<TopicPartition, OffsetAndMetadata> committed =
			consumer.committed(Set.of(new TopicPartition("kbgh1123", 0), new TopicPartition("kbgh1123", 1)));
	System.out.println(in + "\n" + committed);
}
```
```
foo
{kbgh1123-1=OffsetAndMetadata{offset=0, leaderEpoch=null, metadata=''}, kbgh1123-0=OffsetAndMetadata{offset=0, leaderEpoch=null, metadata=''}}
```

garyrussell:
Oh; sorry - I see it's these logs he's complaining about `Found no committed offset for partition kbgh1123-1`.

I still think it would be better to fetch them in one call; why not advise the user the set the log level for `o.a.k.c.c.internals.ConsumerCoordinator` to WARN?

sobychacko:
@garyrussell Ya, we can certainly advise the user of that. I saw your comment on the issue. I was thinking that if we already know that the end offset for the partition is `0`, then why attempt another `committed(...)` call for finding the current offset? But now I realize that, in between those checks, if the offsets have advanced, then we will miss those updates. Therefore, I could see this PR as moot. We can close this PR and then wait for the user to confirm if the `WARN` approach is all they need. What do you think?

garyrussell:
Agreed. But I would still prefer to see a single call to `committed()`. It is a blocking call. Perhaps we should open a new issue for that.

sobychacko:
Closing this PR; The above concern is addressed through https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/1130.

Issue #1126, opened 2021-08-19
Streams Binder does not honor the binding autostartup property


garyrussell:
See https://github.com/spring-projects/spring-kafka/issues/1899#issuecomment-898944583

`StreamsBuilderFactoryBean`s are unconditionally started in `StreamsBuilderFactoryManager.start()`.

The `StreamsBuilderFactoryBean` should not be started if the binding `autoStartup` property is false.

Users can then start the binding manually as with other binders.

PR #1127, opened 2021-08-20
Honoring auto startup in Kafka Streams binder

sobychacko:
When using Kafka Streams bineder, the processors are started
unconditionally, i.e. autoStartup is always true by default.
If spring.kafka.streams.auto-startup is set, then honor that
as the auto-startup flag in the binder.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1126

### DIFF
--- a/docs/src/main/asciidoc/kafka-streams.adoc
+++ b/docs/src/main/asciidoc/kafka-streams.adoc
@@ -1758,6 +1758,30 @@ For instance, if we want to change the header key on this binding to `my_event` 
 
 `spring.cloud.stream.kafka.streams.bindings.process-in-0.consumer.eventTypeHeaderKey=my_event`.
 
+=== Manually starting Kafka Streams processors
+
+Spring Cloud Stream Kafka Streasm binder offers an abstraction called `StreasmBuilderFactoryManager` on top of the `StreamsBuilderFactoryBean` from Spring for Apache Kafka.
+This manager API is used for controlling the multiple `StreamsBuilderFactoryBean` per processor in a binder based application.
+Therefore, when using the binder, if you manually want to control the auto starting of the various `StreamsBuilderFactoryBean` objects in the application, you need to use `StreamsBuilderFactoryManager`.
+You can use the property `spring.kafka.streams.auto-startup` and set this to `false` in order to turn off auto starting of the processors.
+Then, in the application, you can use something as below to start the processors using `StreamsBuilderFactoryManager`.
+
+```
+@Bean
+public ApplicationRunner runner(StreamsBuilderFactoryManager sbfm) {
+    return args -> {
+        sbfm.start();
+    };
+}
+```
+
+This feature is handy, when you want your application to start in the main thread and let Kafka Streams processors start separately.
+For example, when you have a large state store that needs to be restored, if the processors are started normally as is the default case, this may block your application to start.
+If you are using some sort of livenes probe mechanism, it may think that the application is down and attempt a restart.
+In order to correct this, you can set `spring.kafka.streams.auto-startup` to `false` and follow the approach above.
+
+Keep in mind that, when using the Spring Cloud Stream binder, you are not directly dealing with `StreamsBuilderFactoryBean` from Spring for Apache Kafka, rather `StreasmBuilderFactoryManager`, as the `StreamsBuilderFactoryBean` objects are internally managed by the binder.
+
 === Binding visualization and control in Kafka Streams binder
 
 Starting with version 3.1.2, Kafka Streams binder supports binding visualization and control.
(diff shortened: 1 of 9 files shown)


garyrussell (changes requested):
This is useful, but doesn't really address the issue completely.

Individual bindings have an `autoStartup` property; it is being ignored.

https://docs.spring.io/spring-cloud-stream/docs/3.1.3/reference/html/spring-cloud-stream.html#_consumer_properties

I envisioned something like this in `SBFM.start()`...

```java
// Starting the stream.
if (consumerPropertiesForThisBinding(streamsBuilderFactoryBean).isAutoStartup()) {
	streamsBuilderFactoryBean.start();
}
```
I can envision a scenario where they want some streams to auto start and others to not.

Individual bindings that are not autoStartup can be manually started, as described here https://docs.spring.io/spring-cloud-stream/docs/3.1.3/reference/html/spring-cloud-stream.html#binding_visualization_control

This would align the stream binding lifecyle support with the other binders.

I am not saying tear down what you have done; I think it is valid, but I also think we need to manage each binding separately.

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1870,6 +1870,101 @@ When there are multiple bindings present on a single function, invoking these op
...
+public BiFunction<KStream<?, ?>, KTable<?, ?>, KStream<?, ?>> process3() {
+
+}
+
+```
+
+In this scenario above, if you set `spring.kafka.streams.auto-startup` to `false`, then none of the processors will auto start during the application startup.
+In that case, you have to programmatically start them as described above by calling `start()` on the underlying `StreamsBuilderFactoryManager`.
+However, if we have a use case to selectively disable only one processor, then you have to set `auto-startup` on the individual binding for that processor.
+Let us assume that we don't want our `process3` function to not auto start.
```
garyrussell, line 1925:
```suggestion
Let us assume that we don't want our `process3` function to auto start.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1870,6 +1870,101 @@ When there are multiple bindings present on a single function, invoking these op
...
+This is a `BiFunction` with two input bindings - `process3-in-0` and `process3-in-1`.
+In order to avoid auto start for this processor, you can pick any of these input bindings and set `auto-startup` on them.
+It does not matter which binding you pick; if you wish, you can set `auto-startup` to `false` on both of them, but one will be sufficient.
+This is because, both of these bindings share the same `StreamsBuilderFactoryBean` behind the scenes by the binder.
```
garyrussell, line 1929: I am not sure we should encourage this. Maybe

Because they share the same factory bean, you don't have to set autoStartup to false on both bindings, but it probably makes sense to do so, for clarity.

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1870,6 +1870,101 @@ When there are multiple bindings present on a single function, invoking these op
...
+@Autowired
+BindingsEndpoint endpoint;
+
+@Bean
+public ApplicationRunner runner() {
+    return args -> {
+        endpoint.changeState("process3-in-0", State.STARTED);
+    };
+}
+
```
garyrussell, line 1962: Also add a link to the section that describes this mechanism.

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/StreamsBuilderFactoryManager.java
```diff
@@ -81,19 +89,24 @@ public synchronized void start() {
...
 					// Starting the stream.
-					streamsBuilderFactoryBean.start();
-					this.kafkaStreamsRegistry.registerKafkaStreams(streamsBuilderFactoryBean);
+					final Map<StreamsBuilderFactoryBean, List<ConsumerProperties>> bindingServicePropertiesPerSbfb =
+							this.kafkaStreamsBindingInformationCatalogue.getConsumerPropertiesPerSbfb();
+					final List<ConsumerProperties> consumerProperties = bindingServicePropertiesPerSbfb.get(streamsBuilderFactoryBean);
+					final boolean autoStartupDisabledOnAtLeastOneConsumerBinding = consumerProperties.stream().anyMatch(consumerProperties1 -> !consumerProperties1.isAutoStartup());
+					if (!autoStartupDisabledOnAtLeastOneConsumerBinding) {
+						streamsBuilderFactoryBean.start();
+						this.kafkaStreamsRegistry.registerKafkaStreams(streamsBuilderFactoryBean);
```
garyrussell, line 108: Don't we still need to register it, even if it is not started?

garyrussell (changes requested):

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/StreamsBuilderFactoryManager.java
sobychacko: We will get the undelrying `KafkaStreams` object only if we start the `StreamsBuilder`. Each time we stop/start, it is a new `KafkaStreams` object.

garyrussell: Oh; I see that registration/unregistration is done in the binding start/stop too and it is idempotent since they are held in a set.

However, that set needs to be a `ConcurrentHashMap.newKeySet()` because the bindings can be started/stopped on different threads.

garyrussell (approved):

Issue #1065, opened 2021-04-28
Kafka Streams Binder error using interceptor.classes for Consumer or Producer


ceccone:
I've found a problem upgrading to Spring Cloud Hoxton.SR11 (and 2020.0.2).

When using `spring.cloud.stream.kafka.streams.binder.`
`consumerProperties.interceptor.classes=xxx`
or 
`producerProperties.interceptor.classes=xxx`
with Kafka Streams Binder a `java.lang.ClassCastException` is thrown.

The changes made for #997 in `KafkaStreamsBinderSupportAutoConfiguration` are putting both `interceptor.classes` entries in the global properties map, and I suspect that's the problem.

The producer property overrides the consumer one, when creating the kafka consumer it tries to cast the producer interceptor to `org.apache.kafka.clients.consumer.ConsumerInterceptor` throwing a ClassCastException.  
If I only use a consumer interceptor, eventually when creating the kafka producer it reads the global `interceptor.classes` entry  and tries to cast the consumer interceptor to `org.apache.kafka.clients.producer.ProducerInterceptor` also throwing a ClassCastException.

I've made a change to `kafka-streams-word-count` sample project to reproduce the problem: [kafka-streams-word-count](https://github.com/ceccone/spring-cloud-stream-samples/tree/master/kafka-streams-samples/kafka-streams-word-count).
Just run the test class `KafkaStreamsWordCountApplicationTests` to reproduce the problem.

garyrussell:
This has already been discussed here https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/977

It's a KafkaStreams problem; I added a work around there.

garyrussell:
This has already been discussed here https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/977

It's a KafkaStreams problem.

garyrussell:
As a work around, you can implement a single interceptor that implements both interfaces.

ceccone:
Thanks for the feedback @garyrussell.

I was digging into KafkaStreams source code and issue tracker to find the problem and if it was ever reported.

Then, I came across [org.apache.kafka.streams.StreamsConfig](https://kafka.apache.org/25/javadoc/index.html?org/apache/kafka/streams/StreamsConfig.html), which says that we can use prefixes to isolate consumer/producer configs from other client configs.
For instance, we have these options there: 
```java
public static final String CONSUMER_PREFIX = "consumer.";
[...]

PR #1131, opened 2021-08-25
Consumer/Producer prefix in Kafka Streams binder

sobychacko:
Kafka Streams allows the applications to separate the consumer and producer
properties using consumer/producer prefixes. Add these prefixes automatically
if they are missing from the application.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1065

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
@@ -98,6 +98,9 @@ public class KafkaStreamsBinderSupportAutoConfiguration {
 
 	private static final String GLOBALKTABLE_BINDER_TYPE = "globalktable";
 
+	private static final String CONSUMER_PROPERTIES_CONSUMER_PREFIX = "consumer.";
+	private static final String PRODUCER_PROPERTIES_CONSUMER_PREFIX = "producer.";
+
 	@Bean
 	@ConfigurationProperties(prefix = "spring.cloud.stream.kafka.streams.binder")
 	public KafkaStreamsBinderConfigurationProperties binderConfigurationProperties(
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
```diff
@@ -98,6 +98,9 @@
 
 	private static final String GLOBALKTABLE_BINDER_TYPE = "globalktable";
 
+	private static final String CONSUMER_PROPERTIES_CONSUMER_PREFIX = "consumer.";
+	private static final String PRODUCER_PROPERTIES_CONSUMER_PREFIX = "producer.";
```
garyrussell, line 102:
```suggestion
	private static final String CONSUMER_PROPERTIES_PREFIX = "consumer.";
	private static final String PRODUCER_PROPERTIES_PREFIX = "producer.";
```

spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java
```diff
@@ -282,6 +286,18 @@ else if (configProperties
 				Collectors.toMap((e) -> String.valueOf(e.getKey()), Map.Entry::getValue));
 	}
 
+	private void addPrefix(Properties properties, Map<String, Object> mergedConsProdConfig, String prefix) {
+		Map<String, Object> mergedConfigs = new HashMap<>();
+		for (String key : mergedConsProdConfig.keySet()) {
+			if (!key.startsWith(prefix)) {
+				mergedConfigs.put(prefix + key, mergedConsProdConfig.get(key));
+			}
```
garyrussell, line 294:
```suggestion
			mergedConfig.put(key.startsWith(prefix) ? key : prefix + key, mergedConsProdConfig.get(key));
```

Also need to add if already prefixed.

garyrussell (changes requested):

Issue #1135, opened 2021-08-30
Disable Container Retries (by default) When Binding Retries are Enabled Without EnableDlq


garyrussell:
Recent versions of S-K configure a `SeekToCurrentErrorHandler` with 9 retries (10 delivery attempts), previously only a logging handler was configured.

If retries are enabled in the binder, with no DLQ, we get compounded retries (30 by default).

In this case, set the backoff in the STCEH to `new FixedBackOff(0L, 0L)`.

PR #1136, opened 2021-08-31
GH-1135: Disable container retries when no DLQ set

sobychacko:
Disable default container retries when binding retries are enabled
(maxAttempts > 1) and no DLQ set.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1135

### DIFF
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -748,11 +748,11 @@ public class KafkaMessageChannelBinder extends
 		else if (!extendedConsumerProperties.isBatchMode()
 				&& extendedConsumerProperties.getMaxAttempts() > 1
 				&& transMan == null) {
-
 			kafkaMessageDrivenChannelAdapter
 					.setRetryTemplate(buildRetryTemplate(extendedConsumerProperties));
 			kafkaMessageDrivenChannelAdapter
 					.setRecoveryCallback(errorInfrastructure.getRecoverer());
+			messageListenerContainer.setCommonErrorHandler(new DefaultErrorHandler(new FixedBackOff(0L, 0L)));
 		}
 		else if (!extendedConsumerProperties.isBatchMode() && transMan != null) {
 			messageListenerContainer.setAfterRollbackProcessor(new DefaultAfterRollbackProcessor<>(
--- a/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -107,6 +107,7 @@ import org.springframework.kafka.listener.ConsumerAwareRebalanceListener;
 import org.springframework.kafka.listener.ConsumerProperties;
 import org.springframework.kafka.listener.ContainerProperties;
 import org.springframework.kafka.listener.DefaultAfterRollbackProcessor;
+import org.springframework.kafka.listener.DefaultErrorHandler;
 import org.springframework.kafka.support.Acknowledgment;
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -733,7 +734,18 @@ else if (applicationContext != null) {
...
+				messageListenerContainer.setAfterRollbackProcessor(new DefaultAfterRollbackProcessor<>(new FixedBackOff(extendedConsumerProperties.getBackOffInitialInterval(),
+						extendedConsumerProperties.getMaxAttempts() - 1)));
+			}
+			else {
+				messageListenerContainer.setCommonErrorHandler(new DefaultErrorHandler(new FixedBackOff(extendedConsumerProperties.getBackOffInitialInterval(),
+						extendedConsumerProperties.getMaxAttempts() - 1)));
+			}
+		}
+		else if (!extendedConsumerProperties.isBatchMode()
 				&& extendedConsumerProperties.getMaxAttempts() > 1
```
garyrussell, line 749: No; this will still create a compound retry (3x3 by default).

We need `new FixedBackoff(0L, 0L)` until we remove the `RetryTemplate` altogether.

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -1298,6 +1298,55 @@ public void testDefaultAutoCommitOnErrorWithDlq() throws Exception {
...
+		assertThat(handledMessage).isNotNull();
+		assertThat(
+				new String((byte[]) handledMessage.getPayload(), StandardCharsets.UTF_8))
+				.isEqualTo(testMessagePayload);
+		assertThat(handler.getInvocationCount())
+				.isEqualTo(consumerProperties.getMaxAttempts());
+		binderBindUnbindLatency();
+		consumerBinding.unbind();
+		producerBinding.unbind();
+	}
```
garyrussell, line 1348: It's not clear what this is testing.

garyrussell (changes requested):

spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
garyrussell: Does this work now? It still looks to me like you are setting `maxAttempts` in both the `RetryTemplate` and `STCEH/DARP`.

sobychacko: In the case where DLQ is not set and maxAttempts > 1, we are not getting into the conditional branch where we set the `RetryTemplate`. Do you see that is still getting applied somehow?

sobychacko: In the test, there is no DLQ and it verifies that the handler is invoked only as many times as the `maxAttempts` set through the binding.

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
garyrussell: I didn't look closely enough. This new `if()` block is the problem.

It prevents the `RetryTemplate` from being setup (in the next block).

I see that I already dealt with transactions in the `else` - disabling RetryTemplate and constructs an `ExponentialBackOff` from the binding properties.
(We can't use a `RetryTemplate` with transactions because it would run in the transaction, we need the retries outside of the transaction.

I believe you should remove the new if block and set the STCEH with a `FixedBackOff(0L, 0L)` in the next block (after setting the `RetryTemplate`.

garyrussell (approved):

PR #1151, opened 2021-09-23
Add default mappings provider for Kafka Streams

bono007:
Adds the default mappings provider for Kafka Streams so that the core SCS BindAdvise picks up the defaults for the KafkaStreamsExtendedBindingProperties. 

Fixes gh-1152

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/ExtendedBindingHandlerMappingsProviderAutoConfiguration.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/ExtendedBindingHandlerMappingsProviderAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 the original author or authors.
+ * Copyright 2018-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/ExtendedBindingHandlerMappingsProviderAutoConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+@Configuration(proxyBeanMethods = false)
+public class ExtendedBindingHandlerMappingsProviderAutoConfiguration {
+
+	@Bean
+	public MappingsProvider kafkaStreamsExtendedPropertiesDefaultMappingsProvider() {
+		return () -> {
+			Map<ConfigurationPropertyName, ConfigurationPropertyName> mappings = new HashMap<>();
+			mappings.put(
+					ConfigurationPropertyName.of("spring.cloud.stream.kafka.streams"),
```
bono007, line 42: I moved this from the Kafka binder mapping provider. Does this get used by anything?

spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/ExtendedBindingHandlerMappingsProviderAutoConfigurationTests.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.cloud.stream.binder.kafka.streams.properties.KafkaStreamsExtendedBindingProperties;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link ExtendedBindingHandlerMappingsProviderAutoConfiguration}.
+ */
+class ExtendedBindingHandlerMappingsProviderAutoConfigurationTests {
+
+	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
```
bono007, line 32: There are enough moving pieces (required configs) to get the bind advise wired up properly that it was cleaner to simply run an auto-configured dummy app which does the trick.

I also ran this w/ the auto-config disabled (removed from spring.factories) and watched the 2nd test fail (mimics the bug report).

spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/ExtendedBindingHandlerMappingsProviderConfiguration.java
```diff
@@ -40,10 +40,6 @@ public MappingsProvider kafkaExtendedPropertiesDefaultMappingsProvider() {
 			mappings.put(
 					ConfigurationPropertyName.of("spring.cloud.stream.kafka.bindings"),
 					ConfigurationPropertyName.of("spring.cloud.stream.kafka.default"));
-			mappings.put(
```
bono007, line 43: Again, I moved this to the new default mappings provider. Let me know if I am missing something and this is actually being used in the kafka binder module.

sobychacko:
@bono007 Thanks for the PR. It makes sense. The original `ExtendedBindingHandlerMappingsProviderConfiguration` is only bootstrapped when we have Kafka binder in the classpath. If an application only requires Kafka Streams binder, then we never got this through auto configuration. Your changes address that. I will review it further and test it on my end. Then will merge.

sobychacko (approved):

sobychacko:
PR merged upstream.

Issue #1161, opened 2021-10-19
InteractiveQueryService breaks Streams Application Isolation if mutliple Stream Apps using same Store Name


nico-javadev:
Hi there,
not sure if this is a really a Bug or if some other principle applies that should avoid this situation, but here's an interesting detail that we found out about the behavior of InteractiveQueryService, which produces an unwanted behavior in our use case:

When running multiple Stream Apps in the same JVM where the same State Store name is used in the different Streaming Apps, the InteractiveQueryService does not respect the StreamApp origin of the StateStore to query, but just returns the first Store with that name and type from any of the known KafkaStreams instances within the KafkaStreamsRegistry.

In our case this causes errors in processing, since we're using the same Store Names for our Query Services (as a common lib) but require each Stream App to get their specific instance of it's store, even if deployed in the same unit.

The [code responsible](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/f3a954fad71c3fc87c87bd98d0e66735ba9b766f/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java#L100) is pretty straight forward, so I would like to know if this behavior is on purpose, or if this can be considered a bug?

To avoid this, in our case unwanted, behavior we patched the InteractiveQueryService to filter the KafkaStreams instances by matching the Thread's name (which contains the StreamApp's application id). But this is a not so clean workaround I would say, so not gonna share that in the first place here ;)

I'm happy to share more details about the Use Case and workaround if required, but first I would like to know your opinion on this topic and if maybe there's a totally different approach to this.

sobychacko:
@nico-javadev Thanks for this report. It looks like you have an application with multiple Kaka Streams functions (i.e. separate topologies with their own `applicaiton.id`). Is that a true assumption? If that is true, the lack of a guard when querying the KafkaStreams instances from the registry is a bug I think. You are welcome to contribute a PR fixing this.

nico-javadev:
Thanks, yes your assumption is correct. I will prepare a PR :)

sobychacko:
@nico-javadev I see that you have some changes in a branch. If the work is done, could you submit it as a PR?

nico-javadev:
PR is open. Not 100% happy with it since I would consider the Thread name matching rather a hack, but didn't come up with a smarter solution and works for our implementation for now :)
[...]

PR #1166, opened 2021-11-08
PR Issue #1161

### LINKED ISSUE #1167: Question: how to send messages to multiple topics?
I want to send messages to multiple topics at once (no branching), how can I do this with spring cloud stream?

nico-javadev:
This PR Safe Guards State Store Instances in case there are multiple KafkaStreams instances present that have distinct application IDs but share State Store Names.

Change is backwards compatible: in case no KafkaStreams association of the Thread can be found, all Local State Stores are queried as before.
In case an associated KafkaStreams Instance is found, but required StateStore is not found in this instance, a Warning is issued but backwards compatibility is preserved by looking up all State Stores.

Some thoughts / downsides regarding implementation:
- is there a better way to identify if a Thread was created by a Streams instance then to match the Threads name?
- in case a State Store from another Streams Instance is used globally on purpose (e.g. saving space), many Warnings will appear now. Should we introduce a binder property maybe to clearly allow a Streams Application to make use of other's Applications State Stores and disable the Warning in this case?
- I didn't manage to extend the Test Case with a valid scenario since im not very familiar with the way the bindings are created here. But the code is included into our app as Hotfix and works :D 

Fixes #1167

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -14,9 +14,17 @@ Edit the files in the src/main/asciidoc/ directory instead.
 
 image::https://circleci.com/gh/spring-cloud/spring-cloud-stream-binder-kafka.svg?style=svg["CircleCI", link="https://circleci.com/gh/spring-cloud/spring-cloud-stream-binder-kafka"]
 image::https://codecov.io/gh/spring-cloud/spring-cloud-stream-binder-kafka/branch/{github-tag}/graph/badge.svg["codecov", link="https://codecov.io/gh/spring-cloud/spring-cloud-stream-binder-kafka"]
-image::https://badges.gitter.im/spring-cloud/spring-cloud-stream-binder-kafka.svg[Gitter, link="https://gitter.im/spring-cloud/spring-cloud-stream-binder-kafka?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge"]
+image::https://badges.gitter.im/spring-cloud/spring-cloud-stream.svg[Gitter, link="https://gitter.im/spring-cloud/spring-cloud-stream?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge"]
 
 // ======================================================================================
+
+//= Overview
+[partintro]
+--
+This guide describes the Apache Kafka implementation of the Spring Cloud Stream Binder.
+It contains information about its design, usage, and configuration options, as well as information on how the Stream Cloud Stream concepts map onto Apache Kafka specific constructs.
+In addition, this guide explains the Kafka Streams binding capabilities of Spring Cloud Stream.
+--
 
 == Apache Kafka Binder
 
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -7,7 +7,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-stream-binder-kafka-parent</artifactId>
-		<version>3.2.0-M3</version>
+		<version>3.2.0-SNAPSHOT</version>
 	</parent>
 	<packaging>jar</packaging>
 	<name>spring-cloud-stream-binder-kafka-docs</name>
--- a/docs/src/main/asciidoc/_configprops.adoc
+++ b/docs/src/main/asciidoc/_configprops.adoc
@@ -9,6 +9,7 @@
 |spring.cloud.stream.dynamic-destinations | `[]` | A list of destinations that can be bound dynamically. If set, only listed destinations can be bound.
 |spring.cloud.stream.function.batch-mode | `false` | 
 |spring.cloud.stream.function.bindings |  | 
+|spring.cloud.stream.function.definition |  | Definition of functions to bind. If several functions need to be composed into one, use pipes (e.g., 'fooFunc\|barFunc')
 |spring.cloud.stream.instance-count | `1` | The number of deployed instances of an application. Default: 1. NOTE: Could also be managed per individual binding "spring.cloud.stream.bindings.foo.consumer.instance-count" where 'foo' is the name of the binding.
 |spring.cloud.stream.instance-index | `0` | The instance id of the application: a number from 0 to instanceCount-1. Used for partitioning and with Kafka. NOTE: Could also be managed per individual binding "spring.cloud.stream.bindings.foo.consumer.instance-index" where 'foo' is the name of the binding.
 |spring.cloud.stream.instance-index-list |  | A list of instance id's from 0 to instanceCount-1. Used for partitioning and with Kafka. NOTE: Could also be managed per individual binding "spring.cloud.stream.bindings.foo.consumer.instance-index-list" where 'foo' is the name of the binding. This setting will override the one set in 'spring.cloud.stream.instance-index'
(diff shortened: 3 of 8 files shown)


nico-javadev:
PR fixes https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1161

sobychacko:
@nico-javadev Thanks for this PR! It looks great. I have a question for you. What do you think about making this thread context check as an opt-in feature? For example, we can keep the current behavior as it is, but if a new binder property is enabled, then we will switch to isolate retrieving the state store based on the calling thread for the `KafkaStreams`. The reason why I think this way is that most often, a user app might contain a single `KafkaStreams` with a dedicated state store. While the use-case you are addressing is a very valid one, I wonder if this is an edge case. By making this optional, we will still support both cases, but for the per-thread check to work, users have to enable it. What do you say? If you strongly feel otherwise, then we can certainly consider that too. This way, the current behavior continues while addressing the issue with all those warnings you mentioned.  The way you retrieve the thread name looks good to me. Let's go with that and we can always improve later. 
Let us know what you think about these comments.

nico-javadev:
@sobychacko : Sounds good. My proposal was to introduce a binder property the other way around, but I'm fine with activating the isolation per property. 
With a specific property (e.g. stateStoreIsolation=true ? ) I would make the Warning an Error though if a Thread is associated with a Streams instance and the State Store to look up doesn't exist in that instance.
I'll adapt the PR and then we'll see :)

nico-javadev:
@sobychacko : I updated the PR but found some issues:

1. a binder property on a global, non-application specific level, wouldn't allow an application specific config of the State Store isolation opt-in. Unfortunately we already face the case, that within one deployment most Stream Apps shall be isolated, but one should explicitly should not be isolated and reuse another Apps store, so we need the config on Streams Application level.

2. So I chose to read a custom StreamsConfig value "statestore.isolation", since this StreamsConfig in our case is the best available place to put the config.

3. This of course causes the Warning "The configuration 'statestore.isolation' was supplied but isn't a known config." which I know from the AdminClient, but isn't too nice to have :D 

Do you see any better place within the Binder scope to achieve the same behavior?
Let ne add a little context on our Use Case cause I know the requirements probably sound a little weird ;) 

- We're building a rather big insurance application based on Event Sourcing with Kafka Streams as base
- We're building flexible deployment Unit's (Spring Boot applications) with multiple or single Kafka Topologies varying with the specific deployment requirements (e.g. all Topologies within one single App for Minimal/Testing Setup or one Topology per App  Deployments with Productive Deployments)
- We're making use of the Streams Binder, but since we have to be more flexible we are instantiating our specific KafkaStreams instances programmatically and extend the properties supplied by the binder (e.g. append applicationId to create unique Streams App)

e.g. here's our custom Provider for a a specific StreamsBuilder to construct the Toplogy for "task expiration checking".
As you can see the StreamsConfig property offers a good place for us to determine the State Store Isolation, but it's of course not good to use a non-defined Streams Config here.

```
public static final String STREAMS_NAME = "task-expiry-streams";
 @Override public String getAppIdExtension() {
        return "-task-expiry";
    }

    @Bean(name = STREAMS_NAME)
    public FactoryBean<StreamsBuilder> streamsBuilder() {
        return buildAndRegisterStreamsFactory(Collections.singletonMap("statestore.isolation","false"));
        //default is 'true' set by global application.yml
    }
```
```
@Autowired
    KafkaStreamsRegistry kafkaStreamsRegistry;

    @Autowired
    KafkaStreamsConfiguration streamsConfig;
    
    @Autowired(required = false)
    CleanupConfig cleanupConfig;

    @Autowired StageNameUtil topicNameUtil;

    protected FactoryBean<StreamsBuilder> buildAndRegisterStreamsFactory(Map<String,Object> additionalProperties) {
        Properties props = streamsConfig.asProperties();
        Map<String, Object> kafkaProps = new HashMap<>();
        props.forEach((k, v) -> kafkaProps.put((String) k, v));
        kafkaProps.put(StreamsConfig.APPLICATION_ID_CONFIG,topicNameUtil.appendStageName(
                props.getProperty(StreamsConfig.APPLICATION_ID_CONFIG) + getAppIdExtension()));
        if(additionalProperties != null){
            kafkaProps.putAll(additionalProperties);
        }
        if(cleanupConfig == null){
            cleanupConfig = new CleanupConfig();
        }
        StreamsBuilderFactoryBean factory = new StreamsBuilderFactoryBean(new KafkaStreamsConfiguration(kafkaProps), cleanupConfig);
        factory.setKafkaStreamsCustomizer(streams -> {
            // Add to Spring Registry to enable InteractiveQueryService
            StreamsHack.registerKafkaStreams(kafkaStreamsRegistry, factory);
        });
        return factory;
```

sobychacko:
Have you considered adding the new property per function [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/main/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/properties/KafkaStreamsBinderConfigurationProperties.java#L153) instead? That will make sure that the isolation occurs per Kafka Streams function in the application. 

For example, you can then provide the property as `spring.cloud.stream.kafka.streams.binder.functions.<function-name>.stateStoreIsolation`.  This way, although you are providing the property through the binder, it will not be global across the binder, but per function (Kafka streams processor) in the app. What do you think about this suggestion? 

You can also consider providing the property on the consumer binding [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/main/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/properties/KafkaStreamsConsumerProperties.java), which again gives you per processor isolation. But, this approach might be a bit lower level than the one I mentioned above (function one).

sobychacko:
@nico-javadev Any updates on my latest comments?

nico-javadev:
Hey @sobychacko ,
unfortunately both options wouldn't fit for our use case.
Thinking about the whole thing again, I realize that we're not really applying anything from the Kafka Streams Binders Features, except some default configs like serdes and configurations and the InteractiveQueryService.

Since we're building a Development Kit for a Workflow Engine System, we want to hide the whole Kafka and Spring Binder level from our end developers and have to be more flexible when it comes to instantiating Stream Topologies and even Stream Applications only based on the Classes loaded at Runtime.

I have committed a small example of how we instantiate our Stream Topologies which shows the way we work with the binder. Unfortunately the test runs into some error, but I hope it shows the case.

So I wonder if what we are doing here has anything to do with the way spring cloud streams binder is intended to be used? I guess not, since we're not applying functions and only use the binder for some basic configuration :D
But since we're not applying the standard logics, I guess our case of distinguishing State Stores per KafkaStreams instance is not really manageable by the binder config.

sobychacko:
Ok, I don't fully understand the issue. However, if you want to go back to your previous commit (before I made my last comment and since it sounded like it worked for you), we can consider merging the PR as a compromise solution for your use case. Please let us know if you want to proceed with that.  If so, please clean up the PR by removing your latest commit.

nico-javadev:
👍  I reverted the PR to the state where:

- store within KafkaStreams instance of Thread is preferred over "foreign" store with same name
- Warning is issued if requested store is not found within KafkaStreams instance of Thread

The main benefit here is to get rid of randomly selecting stores across all KafkaStreams instances in case a store is contained within multiple Streams Instances with same name.

sobychacko:
@nico-javadev One final thing before we merge. Could you please rebase your PR on top of the current main branch? I am getting some merge conflicts while trying to merge. Could you also add your name with the author tag on the two classes you changed?

nico-javadev:
@sobychacko  Done :)

sobychacko:
@nico-javadev Sorry for going back and forth on this. I am still getting merge conflicts and I tried to patch your individual commits, but still see some issues. You may want to rework these commits, make sure that you don't see the above message "This branch cannot be rebased due to conflicts". Oftentimes, that indicates some issues with the individual commits in the PR. After you rework, I should be able to simply rebase your PR branch without any conflicts on top of the latest main at which point I can squash all your commits into a single commit and merge.

sobychacko:
@nico-javadev I see that you closed the PR. I suppose you are planning to send another clean PR?

nico-javadev:
It closed it automatically :D In the progress of cleaning everything up

PR #1169, opened 2021-11-17
#1161 Make InteractiveQueryService respect Streams context

### LINKED ISSUE #1161: InteractiveQueryService breaks Streams Application Isolation if mutliple Stream Apps using same Store Name

### LINKED ISSUE #1167: Question: how to send messages to multiple topics?

nico-javadev:
This PR Safe Guards State Store Instances in case there are multiple KafkaStreams instances present that have distinct application IDs but share State Store Names.

Change is backwards compatible: in case no KafkaStreams association of the Thread can be found, all Local State Stores are queried as before.
In case an associated KafkaStreams Instance is found, but required StateStore is not found in this instance, a Warning is issued but backwards compatibility is preserved by looking up all State Stores.

- store within KafkaStreams instance of Thread is preferred over "foreign" store with same name

- Warning is issued if requested store is not found within KafkaStreams instance of Thread

The main benefit here is to get rid of randomly selecting stores across all KafkaStreams instances in case a store is contained within multiple Streams Instances with same name.

Fixes #1161

### DIFF
--- a/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
+++ b/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
@@ -31,6 +31,7 @@ import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.streams.KafkaStreams;
 import org.apache.kafka.streams.KeyQueryMetadata;
 import org.apache.kafka.streams.StoreQueryParameters;
+import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.errors.InvalidStateStoreException;
 import org.apache.kafka.streams.state.HostInfo;
 import org.apache.kafka.streams.state.QueryableStoreType;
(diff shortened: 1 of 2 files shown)


nico-javadev:
@sobychacko : Sorry for the hazzle with Git. Did some mistake initially when forking the repo which I had to reset now

sobychacko:
Thank you @nico-javadev for re-submitting the PR. Looks good to me and this is now merged upstream, after squashing the commits and some polishing on the commit message. I also backported the changes to the `3.1.x` branch. You are welcome to contribute more PR's!

nico-javadev:
@sobychacko 👍

Issue #1170, opened 2021-11-19
Classpath reference doesn't work for arbitrary consumerProperties


eduardolbueno:
Currently using:

spring-boot 2.5.3
spring-cloud 2020.0.3 (spring-cloud-starter-stream-kafka 3.1.3)

I have a scenario where I need to use a JKS certificate to connect both to Kafka and to the Schema registry. I have declared properties like so:

```
spring:
  cloud:
    stream:
      kafka:
        binder:
          brokers: // omitted
          configuration:
            security:
              protocol: SSL
            ssl:
              truststore:
                type: JKS
                location: classpath:certs/mycert-app.jks
                password: myapp
              keystore:
                type: JKS
                location: classpath:certs/mycert.jks
                password: myapp
          consumerProperties:
            schema:
              registry:
                url: // omitted
                ssl:
                  truststore:
                    location: classpath:certs/mycert-app.jks
                    password: myapp
                  keystore:
                    location: classpath:certs/mycert.jks
                    password: myapp
```

The problem is that the `classpath:` reference only works for the Kafka connection itself. It doesn't work for the Schema, throwing a `java.io.FileNotFoundException: classpath:certs/mycert.jks (No such file or directory)`. 
It works though with an absolute path, like `/Users/eduardolbueno/workspace/myapp/src/main/resources/certs/mycert.jks` but that is subpotimal for cloud environments.
Can I reference the classpath cert in another way?

sobychacko:
@eduardolbueno When you have `classpath:` resources to load for certificates, then you have to provide a property `spring.cloud.stream.kafka.binder.certificateStoreDirectory` with a value of a writable directory on the filesystem. This is because Kafka only allows certificates to be specified as file system resources. When you provide this property, binder copies the certificates to that location first on the file system before loading them. More details are [here](https://docs.spring.io/spring-cloud-stream-binder-kafka/docs/3.1.5/reference/html/spring-cloud-stream-binder-kafka.html#_configuration_options). Search for `certificateStoreDirectory`.

eduardolbueno:
Great, thank you.
I missed it because I was using this documentation as reference: https://cloud.spring.io/spring-cloud-stream-binder-kafka/spring-cloud-stream-binder-kafka.html#_kafka_binder_properties

eduardolbueno:
[...]

PR #1171, opened 2021-11-23
GH-1170: Schema registry certificates

sobychacko:
Move classpath: resources provided as schema registry certificates
into a local file system location.

Adding test and docs.

Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1170

### DIFF
--- a/docs/src/main/asciidoc/overview.adoc
+++ b/docs/src/main/asciidoc/overview.adoc
@@ -151,6 +151,9 @@ Default: `false`.
 
 spring.cloud.stream.kafka.binder.certificateStoreDirectory::
 When the truststore or keystore certificate location is given as a classpath URL (`classpath:...`), the binder copies the resource from the classpath location inside the JAR file to a location on the filesystem.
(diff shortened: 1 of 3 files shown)


garyrussell (commented):
@sobychacko , I have no issue with this solution per se, but...

>I notice that you have provided the configuration under consumerProperties, but I think it might be cleaner to add this support only at the level of binder configuration.

I recently (yesterday) answered a question on Stack Overflow where they wanted to point to a different registry (in spring-kafka) for two different `KafkaTemplate`s, so you might want to reconsider binding Vs. binder configuration for this (or binding overrides of binder properties). In case a user wants to use a different registry for multiple bindings.

https://stackoverflow.com/questions/70028109/spring-kafka-multiple-schema-registries-with-multiple-kafka-templates/70066899#70066899

sobychacko:
@garyrussell That is interesting. I was hoping to centralize this feature only at the binder level, but what you pointed out is a valid concern. I will re-work the PR. Thanks!

garyrussell (approved):

garyrussell:
... and cherry-picked to 3.1.x

PR #1177, opened 2021-12-10
Use extended properties when initializing Consumer and Producer Serde

yanayita:
Addresses #1176

PR #1179, opened 2021-12-20
Fix: KeySerde setup not using expected key type headers

yanayita:
Bug introduced on #1177 , where isKey value is not properly setup when configuring the KeySerde. This leads to the Key Deserializer, using the key type headers for the value


pinguet62 (approved):

sobychacko:
@yanayita Thanks for catching the issue and the PR. I just merged it upstream. Apologies for the delay in reviewing.

PR #1193, opened 2022-01-26
Update documented version of kafka-clients

jaylindquist:
It looks like this has been incorrect for a few versions. The main branch is currently pulling in 3.1.0

https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/main/pom.xml#L26


jaylindquist:
@pivotal-cla This is an Obvious Fix

sobychacko:
Thank you for the PR. Merged upstream.

